PR #2, opened 2015-03-09
A couple of breadcrumb and 404 fixes

xPaw:
- Added a check to see if include exists before trying to find objects in it and renders 404 page if not found
- Breadcrumb improvements so that it actually says where you are now
- Added a panel saying there are no types/functions if include file exists but there is nothing to display


xPaw:
See #3.

PR #3, opened 2015-03-09
docgen updates

xPaw:
This PR replaces crappy search with a proper one that filters the sidebar.

For improved performance, every `name` column should have an index added.

Other changes:
- Added a check to see if include exists before trying to find objects in it and renders 404 page if not found
- Breadcrumb improvements so that it actually says where you are now
- Added a panel saying there are no types/functions if include file exists but there is nothing to display
- Added a label with return type in return value section
- Made custom types clickable
- Added some poor type highlighting in syntax display
- A bunch of minor style related fixes

### DIFF
--- a/exp/docgen/www/index.php
+++ b/exp/docgen/www/index.php
@@ -6,17 +6,35 @@
     
     $Path = isset( $_SERVER[ 'QUERY_STRING' ] ) ? trim( $_SERVER[ 'QUERY_STRING' ], '/' ) : '';
     $RenderLayout = !isset( $_SERVER[ 'HTTP_X_PJAX' ] ) || $_SERVER[ 'HTTP_X_PJAX' ] !== 'true';
-    
-    if (substr($Path, 0, 8) === '__search')
-        $RenderLayout = false;
-    
+
+    // Stupid bugfix when clicking on homepage
+    if (substr($Path, 0, 6) === '_pjax=') {
+        $Path = '';
+    }
+
     if ($RenderLayout) {
-        $CurrentOpenFile = false;
-        $CurrentOpenFunction = false;
+        $Search = false;
         
+        if (substr($Path, 0, 8) === '__search')
+        {
+            $Path = substr($Path, 9);
+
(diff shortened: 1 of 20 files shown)


xPaw:
Also did a couple of fixes with `active` functions highlight in sidebar. @dvander if you'd like, I can squash all my changes.

exp/docgen/www/index.php
```diff
@@ -6,17 +6,35 @@
     
     $Path = isset( $_SERVER[ 'QUERY_STRING' ] ) ? trim( $_SERVER[ 'QUERY_STRING' ], '/' ) : '';
     $RenderLayout = !isset( $_SERVER[ 'HTTP_X_PJAX' ] ) || $_SERVER[ 'HTTP_X_PJAX' ] !== 'true';
-    
-    if (substr($Path, 0, 8) === '__search')
-        $RenderLayout = false;
-    
+
+    // Stupid bugfix when clicking on homepage
+    if (substr($Path, 0, 6) === '_pjax=') {
```
dvander: What exactly is this?

xPaw: Basically, mod_rewrite doesn't pass in path correctly when you try to load homepage using pjax (`/new-api/?_pjax=testest`) resulting in a 404 error. There's most likely a cleaner way of fixing this via `.htaccess`, but I couldn't figure it out.

exp/docgen/www/template/function.php
```diff
@@ -22,41 +45,94 @@
...
-
-<h4 class="sub-header2">Description</h4>
-<?php RenderDescription($PageFunction['brief']); ?>
+<?php if (empty($PageFunction['brief'])) { ?>
+  <p class="text-muted">This function has no description.</p>
+<?php } else { RenderDescription($PageFunction['brief']); } ?>
+
+<?php
+if (isset($Data['return'])) {
+    $Type = NormalizeType($Data['return']['type']);
```
dvander: This is something that should be cross-indexed in the database itself, but that can wait for an eventual docgen rewrite.

PR #4, opened 2015-03-10
Fix #assert and #error using the wrong error numbers

asherkin:
Looks like these got missed when moving the fatal errors to an enum.

@dvander


dvander:
Thanks!

PR #6, opened 2015-03-16
Docgen update 2

xPaw:
Changes:
- Removes hacky rewrite fix I've added in previous PR
- Makes sidebar toggleable on smaller screens
- Nicer 404 page


dvander:
This broke the sidebar when I updated the production site, so I backed it out on master.

xPaw:
Broke how?

xPaw:
I tested it again, and couldn't see how it breaks.

The only thing I can come up with, is that when you deployed it, older cached `style.css` was served, and that breaks the sidebar.

Try re-reverting this and bust the cloudflare cache.

PR #7, opened 2015-03-30
Fix declaring variable on same line after array

peace-maker:
int a[5], b;
b would get a size of |0| on the stack instead of a sizeof(cell_t).
Bug: https://bugs.alliedmods.net/show_bug.cgi?id=6335


dvander:
Great fix. Thanks!

PR #8, opened 2015-03-30
Add missing debug info for multidimensional strings

peace-maker:
```sourcepawn
new String:INVISIBLE[2][8];
```

Multidimensional arrays like this wouldn't get their debug info added to
the debug symbols table.
Bug: https://bugs.alliedmods.net/show_bug.cgi?id=6324

See Fyren's comment in the bug report.
This patch recursively checks the parent symbol of array dimensions
until it finds the top symbol and uses that to check for the compound
level.
This still allows for early exiting the loop when going out of scope.

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -6270,7 +6270,20 @@ static int testsymbols(symbol *root,int level,int testlabs,int testconst)
   int entry=FALSE;
 
   symbol *sym=root->next;
-  while (sym!=NULL && sym->compound>=level) {
+  symbol *parent;
+  while (sym!=NULL) {
+    if (sym->compound < level) {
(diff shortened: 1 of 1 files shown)


dvander:
Looks sane to me. I'm *guessing* that it's okay to just skip to the parent right away and then check the compound level, but this works too.

peace-maker:
The parent is set for enum elements and functions returning an array as well, so i guessed it's better to specifiy it to this special case. Might be a reason those symbols aren't in the debug table.

dvander:
I meant: if (sym->parent && sym->isArray()) .. compound_level = root_symbol(sym)->compound;

I think it would be cleaner, but spcomp is a mess in general.

PR #10, opened 2015-04-04
Fixed build(when building without sourcemod building it)

WildCard65:


Fyren:
What's the issue here?  Syntax for octal in Python 3?  If so, isn't 0o755 what you want and just 755 totally wrong?

WildCard65:
All I know is that python complained about the 5 on the end b4 change.
On 2015-04-10 8:39 PM, "Fyren" <notifications@github.com> wrote:
>
> What's the issue here? Syntax for octal in Python 3? If so, isn't 0o755
what you want and just 755 totally wrong?

PR #12, opened 2015-04-12
Improve build scripts, other small stuff

Fyren:
Adds --build to configure.py, defaults to "all," choices are spcomp, vm, exp, test, and all.

Hooks up --arch in configure.py, which was there before but apparently not used?  Defaults to x86, can also specify x64 (not that there is currently a reason to).

Changes 0755 in a Python script to 0o755 so it works in both 2 and 3.

Adds SM's .gitignore.

Moves checkout-deps.sh into the root since the build directory is ignored.  Updates the Travis config.

SM's script needs a slight tweak because of these changes.  I only tested this on Linux with gcc.

Travis fails to build exp/ because of a clang warning.  Should it be ignored?


testing/generate_scripts.py
```diff
@@ -37,7 +37,7 @@ def main():
         spcomp = args.spcomp,
         objdir = args.objdir,
       ))
-  os.chmod(args.out, 0755)
+  os.chmod(args.out, 0o0755)
```
dvander: 0o0, I see.

dvander:
Thanks for cleaning this up.

PR #13, opened 2015-04-12
Fix a regression when using accessors as |this|.

dvander:
If a |this| was computed off an expression, then it's not actually an l-value. I forgot to do an l-value check before calling rvalue().

For good measure I made a new wrapper around rvalue() that also resets l-value bits when evaluating an accessor.


psychonic:
:goberserk:

PR #14, opened 2015-04-12
Fixed build for SM

WildCard65:
#12 kind of broke build for sm(on my computer)


Fyren:
SM is not intended to build with SP's tip, it's supposed to build against the commit in the submodule.  When it's updated, the SM build scripts will be changed to match.

WildCard65:
IK, but I found an issue with the improved build scripts when I went to
test them against SM.

PR #18, opened 2015-06-15
Fix handling of escaped hex chars \xHH

peace-maker:
\x2A in Strings and chars wouldn't be replaced by the corresponding character *.


dvander:
Great catch, thanks!

PR #20, opened 2015-07-12
Refactor native generation and place restrictions on the native API.

dvander:
This patch does a few things. First, it removes access to the native function pointer from the SourcePawn API. This allows us to introduce new signatures without breaking embedders.

Second, it refactors exit frames to not depend on a native index. Now there is an explicit "FrameType" enumeration. The native index is now stored explicitly in JitExitFrameForLegacyNative. This lets us introduce new exit frame types in the future.

Third, the native invocation helpers have been removed. These had two functions, one, they verified that no stack/heap leaks occurred, and second, they verified that the native was bound. The check for whether or not the native is bound has been inlined directly into JIT code, and the stack/heap leak checks have been removed.

These checks were not very important. SP is implicitly unwound upon returning to the JIT, since the JIT maintains a local copy. HP was not, but now it is explicitly saved and restored.

### DIFF
--- a/include/smx/smx-v2.h
+++ b/include/smx/smx-v2.h
@@ -28,7 +28,7 @@
 #ifndef _INCLUDE_SPFILE_HEADERS_v2_H
 #define _INCLUDE_SPFILE_HEADERS_v2_H
 
-#include <smx-headers.h>
+#include <smx/smx-headers.h>
 
 namespace sp {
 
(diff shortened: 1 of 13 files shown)


dvander:
On my invoke-empty-native benchmark:
```
native void donothing();

public main()
{
  for (int i = 0; i < 100000000; i++)
    donothing();
}
```

Time goes from 540ms to 382ms, which is about a 30% improvement to the raw overhead of invoking a native. In the current JIT design we're pretty much hitting metal now for no-op natives, so the next step is reducing the overhead for natives that actually do stuff.

dvander:
Pushed via command-line.

PR #24, opened 2015-08-02
Fix complex this-values not being preserved across default arguments.

dvander:
A few methodmap fixes ago, we encountered a problem: spcomp evaluates arguments left-to-right, then reverses their evaluation order. The value of `this` is always computed first, and is never re-ordered. If the value of `this` depends on the value of a register, then argument evaluation can destroy it.

To fix this, we detected when the value of `this` was at risk, and started shuffling it down the stack as each argument evaluated. That worked, except, we forgot to carry that solution over to default values.

This fix comes in two parts. The first part attempts to extract a lot of the argument/stack manipulation out into a separate class. This doesn't change any behavior (I hope). The second part actually fixes the bug.


dvander:
Worked out kinks against sm-master, seems to be okay.

PR #25, opened 2015-08-02
Fix printing JIT opcodes with JIT_SPEW defined

peace-maker:
Fixup the debug spew for recent layout changes.


dvander:
Nice, thanks!

PR #26, opened 2015-08-02
Fix unbound return_reported_error_ label

peace-maker:
If there are loops in the compiled function but no native calls, the
return_reported_error_ label is not bound but used by the throw_timeout_
stub.


dvander:
Nice catch, thanks for diagnosing this. Thr fix is correct, but when "return_reported_error_.used()" is false, it should guarantee that the path has no incoming branches. I would rather move the `throw_timeout_` block above the `return_reported_error_` block to make this true.

PR #27, opened 2015-08-02
Fix VS project generation

peace-maker:
ambuild doesn't like two binaries with the same name when generating VS
project files.

The static library and the executable are both called "spcomp" for the
experimental compiler. Rename the static library to "libspcomp".

### DIFF
--- a/exp/compiler/AMBuilder
+++ b/exp/compiler/AMBuilder
@@ -19,7 +19,7 @@
 import os
 
 ### CLI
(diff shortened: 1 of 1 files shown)


dvander:
Hrm, this will make libspcomp come out as "liblibspcomp.a" in ambuild. I'll poke to see if we can just fix this bug in ambuild.

peace-maker:
I guess it's better to be fixed in ambuild then.

PR #30, opened 2015-08-06
Remove an unneeded assert.

VoiDeD:
Per the short comments on IRC, removing this apparently incorrect assert.


dvander:
Err - sorry, I didn't mean this assert was incorrect. I meant something has gone horribly wrong if it fired. 

During the compilation process we assign natives an index based on a monotonic counter. They are then sorted by that index and that becomes the native table. If their index doesn't match the position in the table, then functions will be calling the wrong native.

VoiDeD:
Oh, yikes. I've filed [a bug](https://bugs.alliedmods.net/show_bug.cgi?id=6401) for this instead, then.

PR #32, opened 2015-08-12
Remove ()= binding syntax.

dvander:
A lot of garbage code and test cases go away with this.


dvander:
Merged via command-line.

PR #43, opened 2015-09-13
Fixed AMBuildScript for build option

WildCard65:


dvander:
Shouldn't this be `getattr(builder.options, 'build')`?

WildCard65:
@dvander IDK, I barely know anything about AMBuild, I just followed with how Configure() gets amtl from options, this change does work as intended though.

dvander:
This clearly does not read from `builder.options.build` anymore, so it doesn't work except that it probably falls through to the last case by accident.

WildCard65:
From my testing, this change does cycle to proper build option instead of
what current behaviour does which is build everything

PR #46, opened 2015-09-18
Additional checks for SMXv1 validation

asherkin:
1. The "out of memory" error could never be triggered on allocation failure due to new throwing instead of returning NULL.
2. Tampering with the SMX container could cause inflate() to crash (initially this was an equality check, but I found a single binary on the forums where the file was 1 byte larger than disksize).

@dvander


dvander:
Interesting, I have no idea what that extra byte could be. This is probably an okay thing to have, it makes the SMX format a lot simpler.

PR #49, opened 2015-11-04
Fix -Wshift-negative-value warning in Clang 3.8

asherkin:
GDB appears to agree that these are functionally identical, although that charmax constant appears to be incorrect...

@dvander


dvander:
:ship:

PR #51, opened 2015-11-11
Add experimental console debugger

peace-maker:
Adds a console debugger within a local "shell" to debug plugins.
It's based on pawndbg, the Pawn debugger, by ITB CompuPhase.

A new exported function is added to grab the debugger interface to avoid changing any existing interface. This helps to keep the interface and structure as flexible as possible until we settle on a design.

``` cpp
IConsoleDebugger *GetConsoleDebugger(int apiVersion);
```

The debugger is disabled completely by default, until it is activated through that interface before any plugin is loaded. (Mapped to a config option in core.cfg in SourceMod)
When enabled, it adds code to break on every dbgbreak plugin opcode. If the plugin is marked as being debugged, it breaks on exceptions and set break points.
Normal execution is paused and the user is prompted with a debug shell to show and change values of variables, step through the plugin, dump plugin memory and walk through the stack trace. General options to manage break points and watched variables are available too.

I've tried to abstract the symbols to support packed and unpacked symbols in old plugin binaries, but that could be done more generally in the future, if we'd add a new smxv2 file format.

The sourcepawn watchdog is set to ignore timeouts while the debugger shell is running, to avoid instant timeouts when continuing execution after working in the debugger shell.

Calling into the debugger to check for breakpoints on every dbgbreak instruction for all plugins is quite expensive. We should only patch in the call when debugging is enabled on the plugin and don't slow down all plugins. But this debugger should only enabled in general on development servers, so performance isn't that much of an issue here.

Basic commands as listed by the `?` command:

```
At the prompt, you can type debug commands. For example, the word "step" is a
command to execute a single line in the source code. The commands that you will
use most frequently may be abbreviated to a single letter: instead of the full
word "step", you can also type the letter "s" followed by the enter key.

Available commands:
    B(ack)T(race)       display the stack trace
    BREAK       set breakpoint at line number or variable name
    CBREAK      remove breakpoint
    CW(atch)    remove a "watchpoint"
    D(isp)      display the value of a variable, list variables
    FILES       list all files that this program is composed off
    F(rame)     Select a frame from the back trace to operate on
    FUNCS       display functions
    G(o)        run program (until breakpoint)
    N(ext)      Run until next line, step over functions
    POS     Show current file and line
    QUIT        exit debugger
    SET     Set a variable to a value
    S(tep)      single step, step into functions
    TYPE        set the "display type" of a symbol
    W(atch)     set a "watchpoint" on a variable
    X       examine plugin memory: x/FMT ADDRESS

    Use "? <command name>" to view more information on a command
```

### DIFF
--- a/vm/interpreter.cpp
+++ b/vm/interpreter.cpp
@@ -1033,6 +1033,13 @@ Interpreter::visitSTRADJUST_PRI()
 }
 
 bool
+Interpreter::visitBREAK()
+{
+  // InvokeDebugger(rt_);
+  return true;
+}
+
+bool
 Interpreter::visitHALT(cell_t value)
 {
   // We don't support this. It's included in the bytestream by default, but it
--- a/vm/interpreter.cpp
+++ b/vm/interpreter.cpp
@@ -22,6 +22,7 @@
 #include "pcode-reader.h"
 #include "runtime-helpers.h"
 #include "watchdog_timer.h"
+#include "console-debugger.h"
 #include <amtl/am-algorithm.h>
 #include <amtl/am-float.h>
 #include <fenv.h>
@@ -1035,7 +1036,14 @@ Interpreter::visitSTRADJUST_PRI()
 bool
 Interpreter::visitBREAK()
 {
-  // InvokeDebugger(rt_);
+  if (!env_->get()->consoledebugger()->IsEnabled())
+    return true;
+  
+  int err = InvokeDebugger(rt_->GetBaseContext());
+  if (err != SP_ERROR_NONE) {
+    cx_->ReportErrorNumber(err);
+    return false;
+  }
   return true;
 }
 
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -76,6 +76,15 @@ class Config(object):
       if builder.options.debug == '1':
           cxx.cflags += ['-g3']
 
+      cxx.defines += [
+        'stricmp=strcasecmp',
+        '_stricmp=strcasecmp',
+        '_snprintf=snprintf',
+        '_vsnprintf=vsnprintf',
+        'HAVE_STDINT_H',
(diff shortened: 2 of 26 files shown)


peace-maker:
Rebased against latest master.

asherkin:
Looks like something went very wrong with the merge here @peace-maker

peace-maker:
Whoops, a rebase is more appropriate.

dvander:
The conclusion I've drawn here is that vm/ is the wrong place for a console debugger. It'd be okay in the SourcePawn under experimental/, but it should not be part of the VM. Instead, it'd be better to figure out the most minimal VM changes possible to make such a debugger work. I say this for two reasons:

(1) I am not prepared to maintain the console debugger, but I am prepared to maintain an API.
(2) The internal changes here are not future proof as they expose too much of the implementation. This could make it really difficult to modify spcomp/smx in the future, and we all know how that works out.

I'll comment in a bit on the API changes that are easy to take from this and the changes that are harder.

vm/interpreter.cpp
```diff
@@ -1032,6 +1033,20 @@ Interpreter::visitSTRADJUST_PRI()
...
 
+bool
+Interpreter::visitBREAK()
+{
+  if (!env_->get()->consoledebugger()->IsEnabled())
+    return true;
+  
+  int err = InvokeDebugger(rt_->GetBaseContext());
+  if (err != SP_ERROR_NONE) {
+    cx_->ReportErrorNumber(err);
```
dvander: In the new error handling model, it's better for the debugger callback to actually report the error on its own, and for the interpreter to check Environment::hasPendingException(). See invokeNative() for an example.

peace-maker:
Instead of trying to merge the full console debugger I'll create seperate PRs for basic debugging support.

PR #52, opened 2015-11-17
Allow blaming an IPF for an error.

Fyren:
This modifies a few interfaces, so I bumped both versions.  I have no idea if both need bumping.

IPluginFunction: adds const char \*DebugName()
IErrorReport: adds int Code() and IPF \*Blame()
IPluginContext: adds BlamePluginError(IPF\*, const char\*, ...)

ScriptedInvoker's existing FullName() was changed to DebugName().  It lost some constness since no other IPluginFunction stuff was.

There'll be another PR for SM to use this instead of ThrowNativeError where applicable and update DebugReporter.

### DIFF
--- a/include/sp_vm_api.h
+++ b/include/sp_vm_api.h
@@ -23,8 +23,8 @@
 #include "sp_vm_types.h"
 
 /** SourcePawn Engine API Versions */
-#define SOURCEPAWN_ENGINE2_API_VERSION 0xA
-#define SOURCEPAWN_API_VERSION   0x020A
+#define SOURCEPAWN_ENGINE2_API_VERSION 0xB
+#define SOURCEPAWN_API_VERSION   0x020B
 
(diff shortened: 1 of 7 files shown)


include/sp_vm_api.h
```diff
@@ -982,6 +995,20 @@ namespace SourcePawn
      * @return          Plugin context.
      */
     virtual IPluginContext *Context() const = 0;
+
+    /**
+     * @brief Return the error code of the error report.
+     *
+     * @return          Integer code.
+     */
+    virtual int Code() const = 0;
```
dvander: If possible, I would drop this from the public API. SM isn't supposed to know about codes anymore, since they're kind of dumb.

dvander:
Looks good, thanks for doing this. :ship: w/ the `Code()` thing addressed.

PR #56, opened 2016-02-25
Validate .tags section.

asherkin:
This breaks a couple of increasingly-common binary modifications to deter Lysis.

I'm thinking we should also be more opinionated here and require *all* the sections we expect to exist, and start warning on load for v1.0 SMX versions (without any debug info) - with the aim to eventually kill those off as well.

Review: @dvander @KyleSanderson
CC: @GoD-Tony @peace-maker


dvander:
:ship:

asherkin:
I'll update the SM submodule after landing some other stuff.

PR #58, opened 2016-04-27
Fix compiler crash on inline arrays in conditions

peace-maker:
Code like:

`if ("") // blabla`
or
`if ({1,2}) // bla`

causes an "array must be indexed" error and crashes when it tries to
print the variable name of the array - because there is none for a
constant inline array.

This patch just prints nothing for the variable name.

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -5926,8 +5926,10 @@ static int test(int label,int parens,int invert)
   if (endtok!=0)
     needtoken(endtok);
   if (ident==iARRAY || ident==iREFARRAY) {
-    const char *ptr=sym->name;
-    error(33,ptr);              /* array must be indexed */
+    if (sym)
+      error(33, sym->name); /* array must be indexed */
+    else
+      error(29);                /* invalid expression */
   } /* if */
   if (ident==iCONSTEXPR) {      /* constant expression */
     int testtype=0;


Fyren:
Other places in spcomp uses "-unknown-" for situations like this.

dvander:
For something like this, we could probably add a better error message entirely. The language doesn't allow these arrays to be indexed in-line, so the message is misleading. When there's no variable name it could just be "array or string literal does not have boolean value" or something like that.

Fyren:
If you change it to index (something like "foo"[0]) then the error you get is "invalid expression."  If you want it to error in the first place it should be something like "indexing array literals is not allowed" since you can't do like {true, false}[0] either.

dvander:
Ah, okay. So indeed by saying "array must be indexed", the compiler is suggesting you do something that cannot be done. Using a different error message entirely when no symname exists is better.

peace-maker:
Changed it to print "array or string literal does not have boolean value" when there is no symbol associated.

Fyren:
I think if you try to do something like `if ({0, 1})` then the existing `invalid expression` error is fine.

peace-maker:
Alright, throwing an "invalid expression" error now instead of a new one.

compiler/sc1.cpp
```diff
@@ -5926,8 +5926,10 @@ static int test(int label,int parens,int invert)
   if (endtok!=0)
     needtoken(endtok);
   if (ident==iARRAY || ident==iREFARRAY) {
-    const char *ptr=sym->name;
-    error(33,ptr);              /* array must be indexed */
+    if (sym)
+      error(33, (const char *)sym->name); /* array must be indexed */
```
dvander: nit: no cast needed here.

peace-maker:
Nits picked here :8ball:

PR #59, opened 2016-06-01
Fix an unused arg warning.

dvander:
This is a quick, surgical fix: A new usage flag, uCALLBACK, which tells doarg to treat callback arguments as if the function were public.


KyleSanderson:
:shipit: after build fix.

PR #60, opened 2016-06-03
Allow capital 'x' in hexadecimal literals

collinsmith:
Added support for case-insensitive 'x' in hexadecimal literals

### DIFF
--- a/compiler/sc2.cpp
+++ b/compiler/sc2.cpp
@@ -573,7 +573,7 @@ static int htoi(cell *val,const unsigned char *curptr)
   ptr=curptr;
   if (!isdigit(*ptr))   /* should start with digit */
     return 0;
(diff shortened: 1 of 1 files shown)


asherkin:
I don't think it's a good idea to add this - multiple syntax variations for a feature are alway confusing, it doesn't look clear (0xA0 vs 0XA0), and I don't think this is commonly supported in other C-like languages - and afaik it's never been asked about before.

dvander:
Agree with @asherkin on this one, unless some super convincing rationale can be offered.

collinsmith:
I understand, it's a really minor thing that I noticed while trying to write a grammar for SourcePawn, since I've never noticed a language that didn't support both representations. According to [ISO/IEC 9899:201x](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf) under section 6.4.4.1, C now supports either representation.

PR #62, opened 2016-06-06
Added support for octal literals

collinsmith:
Added support for octal literals using 0o prefix notation.

Original report:
https://bugs.alliedmods.net/show_bug.cgi?id=6319

### DIFF
--- a/compiler/sc.h
+++ b/compiler/sc.h
@@ -684,6 +684,7 @@ void litadd(cell value);
 void litinsert(cell value,int pos);
 int alphanum(char c);
 int ishex(char c);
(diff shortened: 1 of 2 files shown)


Fyren:
You're missing isoctal's declaration in the header.  Going to hold off merging anyway in case anyone has comments on the syntax.

asherkin:
Not really sure on the value of them, but the `0o` syntax seems good to me and unlikely to bite us later.

collinsmith:
Okay, sounds good. I fixed the problem and squashed the commits.

PR #64, opened 2016-06-12
Make casting a void function illegal, make returning a value from a void function illegal (bug 6492)

Fyren:
Not sure if this is all that's necessary.


WildCard65:
The "return in void function" still produces the tag mismatch warning but AFTER the newly added error 88 in this PR when I was just only expecting error 88.

PR #65, opened 2016-06-20
Allow forward declarations of methodmaps.

Fyren:
This involved a huge amount of flailing around.

You can forward declare a methodmap with just the following:

`methodmap <identifier>;`

If a symbol has `ident` `iMETHODMAP` and `usage` doesn't have `uDEFINE` set, it is considered to be forward declared.  Parsing is a little wonky (https://github.com/alliedmodders/sourcepawn/pull/65/commits/56cea7bf223b6daf8f4df51ff5cd3f15f3afa69f#diff-eac7f772819ea504cece6ddb406c66b3R3658) because if semicolons are not required we have to tell the difference between just `methodmap Foo\n` as a forward declaration and `methodmap Foo\n{` as a definition.

It does compile all of SM's base plugins.  I'll poke it some more to see if I find problems and I'll add a test later.

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -3506,9 +3506,9 @@ methodmap_method_t *parse_method(methodmap_t *map)
   return method;
 }
 
-void declare_methodmap_symbol(methodmap_t* map, bool can_redef)
+symbol* declare_methodmap_symbol(methodmap_t* map, bool can_redef, bool is_forward_decl = false, bool have_forward_decl = false)
 {
-  if (can_redef) {
+  if (can_redef || have_forward_decl) {
     symbol *sym = findglb(map->name);
     if (sym && sym->ident != iMETHODMAP) {
       // We should only hit this on the first pass. Assert really hard that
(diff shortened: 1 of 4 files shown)


dvander:
The following test fails - is that intended?

Code: https://gist.github.com/dvander/32ba5d764f441f7243946c96133b861e
Error message: "(7) : error 105: cannot find method or property B.Value"

peace-maker:
@Fyren what's the status with this one? I think the last input was that since methodmaps piggyback on enums, just doing `enum MethodmapName {}` would do the same thing?

dvander:
This really needs the new parser to be completed. Getting it to work inThis really needs the new parser to be completed. Getting it to work in the two-pass system will be so hacky otherwise. the two-pass system will be so hacky otherwise.

dvander:
Re-posting, now that GitHub's UI seems to be working for me again:

This really needs the new parser to be completed. Getting it to work in the two-pass system will be so hacky otherwise.

PR #66, opened 2016-06-22
Remove half-up rounding in x87 FPU code.

dvander:
In bug 5841 we started using SSE, which implicitly switched our rounding mode on most machines to half-even tie-breaking, rather than half-up. The x87 code is broken anyway (see alliedmodders/amxmodx#376), so let's just simplify it.


IgnacioFDM:
Remember OP_RND_TO_CEIL is broken too (although it isn't inconsistent, it's still broken for values >2^30)

dvander:
@IgnacioFDM Good catch! I'll fix that separately.

PR #69, opened 2016-06-23
Fix stripcom (bug 6424)

Fyren:
Code taken from
https://github.com/compuphase/pawn/blob/6de1822ac1625451bf151159bb0178c09f087bea/compiler/sc2.c#L519-530


dvander:
Test would be nice here, if you feel like doing one feel free to just push as a follow-up.

PR #70, opened 2016-06-28
Fix reported position of bad arguments in methodmap function calls.

Fyren:
The error message for bad arguments reports the bad argument's position.  This is off by one for methodmap calls.


dvander:
`!!aImplicitThis` would be a bool, so this is kind of gross... would it be possible to have something like `const int firstUserArg = aImplicitThis ? 1 : 0` instead?

Fyren:
Oh god, so much more typing.

PR #95, opened 2016-08-26
Fix counting of unspecified final dimensions of arrays.

Fyren:
This change is from upstream.

The old code only considered the first innermost array when counting the size of the final dimension.  This breaks in cases like this:

```
char arr[3][2][] = {
  {"123", "123"},
  {"1234","1234"},
  {"123", "123"}
}
```
The compiler would assume the last dimension was sized one cell/four chars and not indeterminate.  This results in the indirection vectors being wrong.

I'll add some tests before merging.

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -2286,26 +2286,24 @@ static void initials2(int ident,int tag,cell *size,int dim[],int numdim,
           err++;
         } /* if */
       } /* for */
-      if (numdim>1 && dim[numdim-1]==0) {
+      if (numdim>1 && dim[numdim-1]==0 && !errorfound && err==0) {
         /* also look whether, by any chance, all "counted" final dimensions are
-         * the same value; if so, we can store this
+         *  the same value; if so, we can store this
          */
         constvalue *ld=lastdim.next;
-        int d,match;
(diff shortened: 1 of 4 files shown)


Fyren:
Threw in an additional change since sometimes sizeof/cellsof would print junk instead of the array name.  `st` was getting clobbered somewhere along the line.

compiler/sc1.cpp
```diff
@@ -2286,26 +2286,24 @@ static void initials2(int ident,int tag,cell *size,int dim[],int numdim,
...
             break;
-          ld=ld->next;
-        } /* for */
-        if (d==dim[numdim-2])
+          count++;
+        }
+        total=dim[numdim-2];
+        for (d=numdim-3; d>=0; d--)
+          total*=dim[d];
+        if (count>0 && count==total)
```
dvander: Can `count>0` ever be false?

Fyren: I don't think so.  If you do this:

```
int broken[2][2][] = {
  { },
  { },
}
```

It errors just before the context of this patch so skips this path.

Fyren: More to the point, I guess calling initarray will always put something in the lastdim list anyway.

PR #96, opened 2016-08-28
Detect jumps past the end of the function

peace-maker:
It's not allowed to jump outside the function. The end of the function
is only known after it has been completely verified, so add a check for
the highest jump target after we know the end.


dvander:
Good idea, thanks for doing this.

PR #97, opened 2016-08-28
Fix parameter count validation

peace-maker:
Copy & paste error in validation of `sysreq.n` parameter count.
`nparams` is the number of parameters passed, not some address.

The `genarray` validation used the same validation logic, but the
`ndims` are the number of dimensions to be generated instead of an
address.

### DIFF
--- a/vm/method-verifier.cpp
+++ b/vm/method-verifier.cpp
@@ -399,7 +399,7 @@ MethodVerifier::verifyOp(OPCODE op)
     cell_t ndims;
     if (!readCell(&ndims))
       return false;
-    return verifyParamCount(ndims);
+    return verifyDimensionCount(ndims);
   }
 
   case OP_SWITCH:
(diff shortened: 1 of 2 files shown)


vm/method-verifier.cpp
```diff
@@ -531,11 +531,19 @@ MethodVerifier::verifyJumpOffset(cell_t offset)
 }
 
 bool
+MethodVerifier::verifyDimensionCount(cell_t ndims)
+{
+	if (ndims < 0 || ndims > sDIMEN_MAX) {
+      reportError(SP_ERROR_INSTRUCTION_PARAM);
```
KyleSanderson: Missing indentation, and soft tab above.

vm/method-verifier.cpp
```diff
@@ -531,11 +531,19 @@ MethodVerifier::verifyJumpOffset(cell_t offset)
...
+      return false;
+    }
+	return true;
```
KyleSanderson: Ditto on Tab issue.

vm/method-verifier.cpp
```diff
@@ -531,11 +531,19 @@ MethodVerifier::verifyJumpOffset(cell_t offset)
 }
 
 bool
+MethodVerifier::verifyDimensionCount(cell_t ndims)
+{
+	if (ndims < 0 || ndims > sDIMEN_MAX) {
```
dvander: This can be `ndims <= 0` I think, generating a 0-dimension array doesn't make sense?

This is good to merge w/ that and indentation/spacing fixed.

peace-maker:
Right.

vm/method-verifier.cpp
```diff
@@ -531,11 +531,19 @@ MethodVerifier::verifyJumpOffset(cell_t offset)
 }
 
 bool
+MethodVerifier::verifyDimensionCount(cell_t ndims)
+{
+    if (ndims <= 0 || ndims > sDIMEN_MAX) {
+      reportError(SP_ERROR_INSTRUCTION_PARAM);
+      return false;
+    }
+    return true;
```
dvander: Indentation in this function should be 2-space instead of 4. Good to merge w/ that fixed.

peace-maker:
Dmn it!

PR #100, opened 2016-10-26
Fix compilation for GCC6

Ni3znajomy:
Similar to https://github.com/alliedmodders/amxmodx/pull/372

Option ```-Wmisleading-indentation``` has been introduced in GCC 6 and it's enabled by ```-Wall```. Due ```-Werror``` option we're unable to successfully build SourcePawn using GCC.


dvander:
Thanks!

PR #102, opened 2016-10-27
Report the right error code for invalid param count

peace-maker:
There's a specific error number for exceeding the maximum parameter
count. Use it!


dvander:
This is technically an instruction encoding error, param count is a runtime thing. The compiler should not be encoding values that the VM won't accept (or vice versa).

peace-maker:
Alright, nvm.

PR #105, opened 2017-01-20
Disallow |halt| instruction

peace-maker:
The compiler never emits this instruction, so don't allow it in a binary.
@asherkin did a test against all uploaded .smx files in the forum to confirm.


asherkin:
Thanks!

PR #110, opened 2017-04-30
Unify parameter count defines in VM and compiler (bug 6540)

peace-maker:
`sARGS_MAX` becomes `SP_MAX_EXEC_PARAMS` and `sMAXARGS` becomes `SP_MAX_CALL_ARGUMENTS` in the VM as well as the compiler.

The method verifier now checks for the maximum allowed arguments of a call,
which differs from the maximum number of parameters when declaring a function in the compiler.

The present name `SP_MAX_EXEC_PARAMS` doesn't really describe the value, but it would require changes to SourceMod too to make it more descriptive.


dvander:
Thanks!

PR #111, opened 2017-05-04
Fix assert and bad error messaging when using constructors as static methods.

dvander:
See: https://forums.alliedmods.net/showthread.php?t=297004


dvander:
Fyren noted that I used the wrong error number here, so fixing that and merging.

PR #112, opened 2017-05-30
Support building spcomp and spshell with Emscripten

asherkin:
```
source ~/emsdk_portable/emsdk_env.sh
CC=emcc CXX=emcc python3 ../configure.py --build=spcomp,vm --enable-optimize
ambuild
```

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -249,18 +249,18 @@ if builder.parent is None:
   root = Config()
   root.configure()
   sp = SourcePawn(root, root.amtl)
-  build = getattr(builder.options, 'build', None)
-  if build == 'spcomp':
-    sp.BuildSpcomp()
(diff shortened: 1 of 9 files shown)


dvander:
ship: assuming CI works

asherkin:
Getting an assert in fail-invalid-array-3d, but all tests pass with a release build.
`Assertion failed: strtol(ld->name,NULL,16)==count % dim[numdim-2]`

This is probably ready for another look @dvander.

The unaligned path in getCellValue is overkill, afaict only LODB will hit this route and it is because it is using a bit of a hack by offsetting the address and masking the result to read sub-cell-sized values - adding direct getCellValue8 and getCellValue16 accessors would possibly be the saner way forward (assuming all 2-byte reads end up being aligned correctly).

PR #114, opened 2017-06-05
Add failing tests for recently discovered SP bugs

asherkin:
early-compound-assignment will currently fail when the peephole optimizer is enabled.
early-return will currently fail when the JIT is disabled.

Not really mergable until these bugs are fixed, but I figured this was probably a good start.


asherkin:
Looks like the AppVeyor build isn't actually checking if tests pass...

dvander:
Great, thanks! Probably we should split these tests into whatever patches end up fixing them. But let's keep it open so they don't get lost.

asherkin:
These now live in their respective PRs.

PR #116, opened 2017-06-05
Correctly propagate return code when running tests

asherkin:


asherkin:
That failure is from the assert I ran into in #112, the test failure was being masked by the previous error checking behavior in the test scripts.

PR #117, opened 2017-06-05
Stop executing opcodes after hitting a RETN (Fix #115)

### LINKED ISSUE #115: Interpreter incorrectly handling OP_RETN
See test in #114.

Correct output:
```
main.1
inner.1
main.2
```

Broken output:
```
main.1
inner.1
inner.2
inner.3
Exception thrown: Stack went below stack boundary
  [0] test.sp::inner, line 13
  [1] test.sp::main, line 19
```

```
[23:57:24] <dvander> asherkin: my guess is it's dumber than that: we never actually return from the interpreter
[23:58:32] <dvander> there doesn't appear to be any code that says "stop trying to run functions and exit"
```

I'm not sure it is actually that, this dumb patch seems to make everything work as expected, but I haven't tried more complex cases:
```
diff --git a/vm/interpreter.cpp b/vm/interpreter.cpp
index 578c1f6..b71e199 100644
--- a/vm/interpreter.cpp
+++ b/vm/interpreter.cpp
@@ -65,6 +65,13 @@ Interpreter::run()
   while (reader_.more()) {
     if (reader_.peekOpcode() == OP_PROC || reader_.peekOpcode() == OP_ENDPROC)
       break;
+
+    if (reader_.peekOpcode() == OP_RETN) {
+      if (!reader_.visitNext())
+        return false;
+      break;
+    }
+
     if (!reader_.visitNext())
       return false;
   }
```

But this looks really silly and I'm not sure how to factor it in properly.

I can't seem to get the compiler to ever emit OP_ENDPROC (or anywhere in the could where it would), what is it used for?

asherkin:
@dvander still not 100% sure this is correct, but it seems to work and be sane, figured I'd give it a shot.

### DIFF
--- a/tests/basic/early-return.out
+++ b/tests/basic/early-return.out
@@ -1,0 +1,3 @@
+main.1
+inner.1
+main.2
--- a/tests/basic/early-return.sp
+++ b/tests/basic/early-return.sp
@@ -1,0 +1,23 @@
+#include <shell>
+
+void inner() {
+  print("inner.1\n");
+  return;
+  print("inner.2\n");
+  return;
+  print("inner.3\n");
+}
+
+public int main() {
+  print("main.1\n");
+  inner();
+  print("main.2\n");
+  return;
+  print("main.3\n");
+  inner();
+  print("main.4\n");
+  return;
+  print("main.5\n");
+  inner();
+  print("main.6\n");
+}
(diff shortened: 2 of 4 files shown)


dvander:
Looks good, r=me. This is exactly what I was thinking the fix would look like, though for some reason I had mentally put the stop flag on PcodeReader which is the wrong place. 

Can you include the other tests you write in this PR, assuming it makes them pass?

PR #118, opened 2017-06-05
Fix optimizer breaking chained assignments.

Fyren:
Includes test from asherkin's other PR (with a little extra).


dvander:
:ship: thanks for looking into this!

PR #119, opened 2017-06-06
Abort array parsing when encountering an error

asherkin:
Fixes `Assertion failed: strtol(ld->name,NULL,16)==count % dim[numdim-2]` in `fail-invalid-array-3d`.


asherkin:
`initarray` has `assert(errorfound!=NULL && *errorfound==FALSE);` at the start.

Issue #122, opened 2017-06-18
sqlite database creation bug


scorpius2k1:
SourceMod Forums Thread: [https://forums.alliedmods.net/showthread.php?p=2529668#post2529668](https://forums.alliedmods.net/showthread.php?p=2529668#post2529668)

I am using the latest SourceMod, Linux Server, Half-Life 2: DM. This is driving me absolutely crazy! I am simply trying to create a database, and it does, but for some reason, if I use a timer, and then call StrEqual...it makes a SECOND empty database with whatever name is in the first field of "StrEqual"

Here is an image from FTP showing the two files being made:
http://imgur.com/GFSrYC2l.png

Compile the code below to reproduce it; the databases can be found here
addons/sourcemod/data/sqlite

```
#include <sourcemod> 

new Handle:g_hDB = INVALID_HANDLE; 

public OnPluginStart() 
{ 
    InitDB(); 
    CreateTimer(3.0, wtf_is_going_on); 
} 

stock Action:wtf_is_going_on(Handle:timer) 
{ 
    if(StrEqual("it_makes_an_empty_database_named_with_whatever_is_here", "")) {} 
} 

stock InitDB() 
{ 
    new String:error[255]; 
    g_hDB = SQLite_UseDatabase("my_database", error, sizeof(error)); 
} 
```

KyleSanderson:
So, this one is a little amazing and makes sense related to other bugs I've seen (if that indeed compiles). If you change your timer callback to public that should fix everything.

scorpius2k1:
Very strange! That definitely fixed the "issue" KyleSanderson, usually something simple. thank you!

KyleSanderson:
@scorpius2k1 While your issue may be fixed, this is (probably!) a compiler bug that we should probably track/fix. Thanks for filing!

scorpius2k1:
No problem at all, thanks for the help & all the great features with SM!

dvander:
Good catch. This is indeed a compiler bug. It's easy to see what's going wrong by comparing the assembly listing diffs: the code to `StrEqual` is missing when `wtf_is_going_on` is a stock. But the call is still there, so it's jumping into who knows where, probably a random address in the code section.

My guess is we're forgetting to call `markusage` on read-but-not-executed stocks. Indeed, that seems to be the case here [1]. That should probably be `uREAD|uCALLBACK`.

[...]

PR #123, opened 2017-06-20
Fix an issue where stock callbacks might generate incorrect code (bug #122)

dvander:
When using a stock from another stock, where the latter stock is only invoked via a callback, the compiler can skip generating code from the first stock. But it still emits the call, and we end up jumping into the wrong function.

The bug is caused by code we soft deleted a long time ago. When we first imported Pawn as SourcePawn there were a number of questionable features we disabled. One of these was a syntax feature where functions could be invoked without parenthesis - a wildly unreadable syntax extension that had hacks sprinkled all throughout the source code to make it function. At the time we wanted to keep merging from upstream so we just disabled the feature at a pinch point and kept all the supporting code in. We ended up diverging so much from upstream that we stopped taking code.

Anyway, code related to this feature was preventing us from recognizing functions-as-callbacks in the first pass. Specifically, if we would lex a token that is a function, and it was not followed by a parenthesis, it would explicitly throw an error in the first pass. Since stocks must be marked as READ in the first pass, we would not propagate stock usage correctly.

The fix is to just remove all the code to support paren-less calls. I also added a fatal error message for this scenario, since we've had bugs like it before.


Fyren:
```
void foo()
{
        print("sub\n");
}

native void print(const char[] str);

public void main() { }
```
This fails an assert with this patch.  If you comment out the native declaration, it doesn't assert.  With master, it doesn't assert.

`sc3.cpp:2474: int primary(value*): Assertion `sc_status==statFIRST' failed.`

Issue #136, opened 2017-08-03
Tag Mismatch On Char Literal Assignment


Headline22:
Test case:
```
int main()
{
	char a = 'a';
}
```
Warning:
```
SourcePawn Compiler 1.8.0.6014
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2015 AlliedModders LLC

C:\Users\Micha\Desktop\test2.sp(3) : warning 213: tag mismatch
```

IRC discussion:
```
<dvander> you'd have to step through matchtag() to figure out why
<dvander> but asherkin's right
<dvander> char resolves to the deprecated "String" tag, which is a big, big hack
<dvander> that technically only applies to arrays
<dvander> so probably it's just some archaic mess up
<dvander> i think that assignment should be supported though
```

Headline22:
Closed.

PR Accepted: #137

PR #137, opened 2017-08-03
Fix char literal tag mismatch (issue #136)

Headline22:
Not sure if this is an acceptable solution to this issue, but I have tested it and it does work. Let me know.


dvander:
Thanks, this fixes a minimal test case. I'll merge this and check that in separately.

PR #139, opened 2017-09-16
Switch to AMBuild 2.1 API

TheDS:
While attempting to port the SourceMod build scripts to the 2.1 API for the 64-bit port, I realized this needed to be switched first.


asherkin:
aiuc, the 2.1 API doesn't currently include Emscripten support which is a bit of a problem for me, but doesn't block doing this.

PR #140, opened 2017-09-17
Create GetFrameIterator & FreeFrameIterator

Headline22:
In order to create RequestStackFrame in SourceMod, we need an
IFrameIterator from sp. This exposes a FrameIterator object for this
purpose.


dvander:
FrameIterators must be used in a very specific way, which is why we didn't have this exposed already. I think as long as the comments are clear it should be okay though.

include/sp_vm_api.h
```diff
@@ -961,6 +1047,18 @@ namespace SourcePawn
      * as the cause.  
      */
     virtual cell_t BlamePluginError(IPluginFunction *pf, const char *msg, ...) = 0;
+   	
+	/**
+     * @brief Returns an IFrameIterator for the current call stack. Must
+	 * be freed by FreeFrameIterator()
+     */
+    virtual IFrameIterator *GetFrameIterator() = 0;
```
dvander: Spacing in the comment looks corrupt - should be spaces instead of tabs. Let's call this "CreateFrameIterator" and add a comment like: "It is illegal to re-enter the VM while a frame iterator is active. The iterator must be processed and destroyed before continuing to run SourcePawn plugins."

include/sp_vm_api.h
```diff
@@ -961,6 +1047,18 @@ namespace SourcePawn
...
+	/**
+     * @brief Returns an IFrameIterator for the current call stack. Must
+	 * be freed by FreeFrameIterator()
+     */
+    virtual IFrameIterator *GetFrameIterator() = 0;
+
+	/**
+     * @brief Frees a IFrameIterator object. Paired with GetFrameIterator() 
+     */
+    virtual void FreeFrameIterator(IFrameIterator *it) = 0;
```
dvander: Same here re: spacing. Let's call this DestroyFrameIterator.

vm/base-context.cpp
```diff
@@ -236,3 +236,14 @@ BasePluginContext::BlamePluginError(SourcePawn::IPluginFunction *pf, const char
   va_end(ap);
   return 0;
 }
+
+IFrameIterator *BasePluginContext::GetFrameIterator()
```
dvander: Local style is:

```
IFrameIterator *
BasePluginContext::GetFrameIterator()
```

(Return type/modifiers on one line, class/method name on next)

vm/base-context.cpp
```diff
@@ -236,3 +236,14 @@ BasePluginContext::BlamePluginError(SourcePawn::IPluginFunction *pf, const char
...
+{
+  FrameIterator *it = new FrameIterator();
+  return it;
+}
+
+void BasePluginContext::FreeFrameIterator(IFrameIterator *it)
+{
+  delete it;
```
dvander: `IFrameIterator` does not have a virtual destructor, so you have to `static_cast<FrameIterator*>(it)` to delete it.

peace-maker:
You should still bump the version define due to the IPluginContext change.

vm/base-context.cpp
```diff
@@ -236,3 +236,16 @@ BasePluginContext::BlamePluginError(SourcePawn::IPluginFunction *pf, const char
...
+{
+  FrameIterator *it = new FrameIterator();
+  return it;
+}
+
+void
+BasePluginContext::DestroyFrameIterator(IFrameIterator *it)
+{
+  delete static_cast<FrameIterator*>(it);
+}
```
KyleSanderson: nit: newline please

dvander:
Sorry for the delay - looks good!

PR #141, opened 2017-09-18
Allow building more than one arch at a time.

TheDS:


psychonic:
Does this work in practice with msvc?

Unlike GCC or Clang, it has a different compiler executable used per arch. I don't believe that just setting /MACHINE is enough. (Which copy of cl.exe that's in the path is typically determined by which arch's vcvars batch file was run).

TheDS:
You raise a good point because it does not appear to work with MSVC. The original idea was to make it possible to build SourceMod for both archs and package them together. So SP needed the same kind of support. This is what was done in MM:S PR as well.  Now I'm not quite sure what approach to take.

asherkin:
We probably need to have AMBuild2 do the detection of MSVC on the system (like CMake does) rather than require running vcvars* / Build Tools Command Prompt, then it can detect all of the available versions / archs and do selection internally.

TheDS:
I've rebased this with the current master and now throw an exception if you attempt to configure with multiple archs using MSVC. Since there are not yet any 64-bit Source 1 games for Windows, I figure this is the easiest path forward for now. Detecting MSVC a different way in AMBuild is not exactly trivial. In addition, an exception is thrown if you specify an arch other than x86, x86_64, or x64 (where x86_64 and x64 are equivalent).

AMBuildScript
```diff
@@ -22,26 +22,55 @@ import sys
...
   def __init__(self):
     super(Config, self).__init__()
-    self.arch = builder.target.arch
+    self.archs = builder.target.arch.split(',')
 
   def configure(self):
     cxx = builder.DetectCxx()
 
+    if cxx.like('msvc') and len(self.archs) > 1:
+      raise Exception('Building multiple archs with MSVC with not currently supported')
```
peace-maker: English!

TheDS: 🤦‍♂️ Thank you.

AMBuildScript
```diff
@@ -22,26 +22,55 @@ import sys
...
   return os.path.abspath(os.path.normpath(path))
 
+def SetArchFlags(compiler, arch, platform):
+  if compiler.like('gcc'):
+    if arch == 'x86':
+      compiler.cflags += ['-msse']
+    else:
+      compiler.cflags += ['-fPIC']
+
+    if arch == 'x64' or arch == 'x86_64':
```
dvander: If possible, having one representation is cleaner... then you know you only have to check  for one thing everywhere, versus both 'x64' and 'x86_64'.

AMBuildScript
```diff
@@ -22,26 +22,55 @@ import sys
...
+      if msvcarch not in compiler.linkflags:
+        compiler.linkflags += [msvcarch]
+    else:
+      if gccarch not in compiler.cflags:
+        compiler.cflags += [gccarch]
+      if gccarch not in compiler.linkflags:
+        compiler.linkflags += [gccarch]
+
+def AppendArchSuffix(binary, name, arch):
+  if arch == 'x64' or arch == 'x86_64':
```
dvander: (like, here)

TheDS:
@psychonic Unless I'm missing something, only you can update the appveyor settings/script to run the right test files? This is currently failing on there.

PR #146, opened 2017-09-25
Remove funcenum and functag syntax.

dvander:
This syntax involves some of the earliest changes we made to Pawn, and the code is extremely gross. The syntax has been deprecated for almost 3.5 years now, and it's quite trivial to update old code to use `typeset` and `typedef`. There I'm proposing we remove this syntax entirely.

This patch will abort compiling any plugin with `funcenum` or `functag`, and will link the user to a URL on how they can update. Of course, people can always continue to use older compilers.


compiler/sc5-in.scp
```diff
@@ -157,7 +157,7 @@ static const char *errmsg[] = {
 /*133*/  "cannot coerce unrelated object types %s and %s\n",
 /*134*/  "type mismatch (%s and %s)\n",
 /*135*/  "cannot use an object in a multi-tag selector\n",
-/*136*/  "unused\n",
+/*135*/  "unused135\n",
```
asherkin: Did something go wonky here? Two '135's.

compiler/tests/fail-functag.sp
```diff
@@ -0,0 +1,3 @@
+funcenum Gab {
+  public:Action(Handle:handle),
+};
```
Timocop: Same as "fail-funcenum.sp"?

dvander:
Thanks, I've addressed these comments.

compiler/tests/fail-functag.sp
```diff
@@ -0,0 +1 @@
+functag public:Action(Handle:handle);
```
peace-maker: The syntax looks wrong. Not that it really matters since it already fails on `functag`..
Same for the funcenum one.

PR #152, opened 2017-10-03
Remove iREFFUNC.

dvander:
I don't know what this is or whether it ever even existed. It's never used in any way, but there are lots of checks for it. It's possible faluco or I added it before SM, during the initial JIT sprint.

My guess is it's supposed to be analogous to iREFARRAY or iREFERENCE, but the necessity of those is dubious to begin with.


dvander:
(Nope - it's in AMX Mod X too. whatever.)

PR #157, opened 2017-10-07
Add x64 support for fake natives.

TheDS:
Using a fake native on x64 with the interpreter crashes in <tt>SourcePawnEngine::FreePageMemory</tt>. In order to make this work, code needs to be generated for a fake native stub. I've taken some of the existing code from the x64 branch to make this happen.

### DIFF
--- a/vm/environment.h
+++ b/vm/environment.h
@@ -73,11 +73,9 @@ class Environment : public ISourcePawnEnvironment
   // Allocate and free executable memory.
   CodeChunk AllocateCode(size_t size);
 
-#if defined(SP_HAS_JIT) || defined(KE_ARCH_X64)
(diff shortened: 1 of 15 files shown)


vm/environment.h
```diff
@@ -176,7 +176,7 @@ class Environment : public ISourcePawnEnvironment
   bool profiling_enabled_;
 
   ke::AutoPtr<CodeAllocator> code_alloc_;
-#if defined(SP_HAS_JIT)
+#if defined(SP_HAS_JIT) || defined(KE_ARCH_X64)
```
dvander: I think it's okay to just require that all archs support CodeStubs. Then we can take out some of these ifdefs.

PR #159, opened 2017-10-10
Copy function debug name onto stack for crash dumps

asherkin:
This is a useful hack I've had sitting in my queue for a while.

Copy the to-be-invoked function's debug name into the `ScriptedInvoker::Invoke` stack frame so it is always available in minidumps even without heap information.

This will hopefully make it easier to track down reproduction steps for plugin-related crashes.

### DIFF
--- a/vm/scripted-invoker.cpp
+++ b/vm/scripted-invoker.cpp
@@ -294,9 +294,9 @@ ScriptedInvoker::Invoke(cell_t *result)
   /* Make the call if we can */
   if (ok) {
     const char *debugName = this->DebugName();
-    size_t debugNameLength = strlen(debugName) + 1;
-    volatile char *debugNameForCrashDumps = (char *)alloca(debugNameLength);
-    SafeStrcpy((char *)debugNameForCrashDumps, debugNameLength, debugName);
+    size_t debugNameLength = strlen(debugName) + 2;
+    volatile char * volatile debugNameForCrashDumps = (char *)alloca(debugNameLength);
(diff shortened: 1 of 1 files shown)


asherkin:
I've played with a few variations in https://godbolt.org/, I think `alloca` remains the best option all-round - the location and behaviour is a lot more predictable with compiler optimisations, and choosing a reasonable buffer size to avoid wasting stack space in all cases without truncating is a PITA.

KyleSanderson:
http://man7.org/linux/man-pages/man3/alloca.3.html

`There is no error indication if the stack frame cannot be extended.
       (However, after a failed allocation, the program is likely to receive
       a SIGSEGV signal if it attempts to access the unallocated space.)

       On many systems alloca() cannot be used inside the list of arguments
       of a function call, because the stack space reserved by alloca()
       would appear on the stack in the middle of the space for the function
       arguments.`

`This function is not in POSIX.1.

       There is evidence that the alloca() function appeared in 32V, PWB,
       PWB.2, 3BSD, and 4BSD.  There is a man page for it in 4.3BSD.  Linux
       uses the GNU version.`

I would avoid alloca at all costs.

asherkin:
As would just calling the function if it was allocating a large enough local variable... that information isn't exactly useful or relevant.

The DebugName of a function is fairly well constrained (and I believe should be well under the limit reasonable to stack alloc - which MS considers to be 1kB for an individual alloc), it just isn't ideal to allocate the full size required for the 90% of cases where it is much shorter.

PR #161, opened 2017-11-07
Remove then/begin/end tokens.

Fyren:
Currently, you can use `*begin` and `*end` to enclose compound statements in a lot of places as well as use `*then` to not require parentheses around the test for `if` statements.  I have no idea why they have asterisks in the first place.  Perhaps a lazy attempt to remove the syntax?

This patch additionally removes the ability to do `while a == 1 do stmt();`.


dvander:
:ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship:

dvander:
I'm pretty sure that was indeed a lazy attempt. I vaguely recall thinking the lexer would reject * because it would match earlier as just a '*', but of course that's not true.

PR #165, opened 2017-12-19
Add "not" optimization.

Kailo97:
```
if (!func())
```
Gives
```
not
jzer n
```
It has to replace to
```
jnz n
```

### DIFF
--- a/compiler/sc7-in.scp
+++ b/compiler/sc7-in.scp
@@ -848,6 +848,32 @@ static SEQUENCE sequences_cmp[] = {
       "jnz %1!",
     seqsize(2,1) - seqsize(1,1)
   },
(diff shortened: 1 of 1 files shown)


asherkin:
This lgtm (being a lift from upstream), but I'd like @Fyren or @dvander to give this a once over before merging as they're more familiar with this beast and our changes to it.

Kailo97:
About this case (probably not generatable)
```
eq.c.pri 0
not
jzer n1
```
all still worse
```
if (!(g_int[0] == 0))
```
gives
```
const.pri 2228
load.i
zero.alt
eq
not
jzer 3292
```

dvander:
This seems reasonable.

PR #175, opened 2018-02-04
Align JIT HALT implementation with Interpreter and Verifier

### LINKED ISSUE #174: Sleep and Exit
I wrote about this feature in Discord and IRC. And retry here, in issue.
Code like:
```sp
public void OnPluginStart() {
    sleep;
}
```
cause the server to crash.
Accelerator: https://crash.limetech.org/uhqigne5rpie

SM Version:
```
 SourceMod Version Information:
    SourceMod Version: 1.8.0.6036
    SourcePawn Engine: SourcePawn 1.8, jit-x86 (build 1.8.0.6036)
    SourcePawn API: v1 = 4, v2 = 11
    Compiled on: Nov  3 2017 19:06:38
    Built from: https://github.com/alliedmodders/sourcemod/commit/fef5583
    Build ID: 6036:fef5583
    http://www.sourcemod.net/
```

Code with this "feature":
https://github.com/alliedmodders/sourcepawn/blob/a1c9bfb6a1a26ef11d84d6443f844f2e6cd90506/compiler/sc1.cpp#L5434-L5437
https://github.com/alliedmodders/sourcepawn/blob/a1c9bfb6a1a26ef11d84d6443f844f2e6cd90506/compiler/sc1.cpp#L5442-L5445

asherkin:
Next step is removing the code paths that can emit reachable `halt`'s from spcomp (exit, sleep, and assert), but this fixes the crash for now.

For shell use it might be nice to have these work (and just map `halt` through to `SetFailState` for SM), but that is a discussion for the future.

Fixes #174

### DIFF
--- a/vm/x86/jit_x86.cpp
+++ b/vm/x86/jit_x86.cpp
@@ -1082,12 +1082,10 @@ Compiler::visitTRACKER_POP_SETHEAP()
 bool
 Compiler::visitHALT(cell_t value)
 {
-  // This should never be hit.
-  __ align(16);
-  __ movl(pri, value);
-  __ testl(eax, eax);
-  jumpOnError(not_zero);
-  return true;
+  // We don't support this. It's included in the bytestream by default, but it
+  // must be unreachable.
+  reportError(SP_ERROR_INVALID_INSTRUCTION);
+  return false;
 }
 
 bool


dvander:
How was this emitting when the verifier rejects it?

asherkin:
The SP version in SM 1.8-dev doesn't have #105.

dvander:
Hah, okay.

PR #176, opened 2018-02-06
Fix initialization of dynamic char[]

peace-maker:
Initializing a string on the heap using
```
char[] str = "this is a value";
// NOT |char str[] = "this is a value";| which is on the stack
```
wasn't copying the correct string.

Add a test to catch this from now on.

Fixes https://github.com/alliedmodders/sourcemod/issues/760


dvander:
Nice catch!

PR #178, opened 2018-02-07
Test Emscripten build on Travis

asherkin:
This PR adds Emscripten build/test to Travis CI.

Due to incompatibilities with the Travis environment and the prebuilt Emscripten SDK, this is using a built-for-Travis `emsdk` from the Urho3D project - which I trust.

The Emscripten builds are fairly slow (especially for debug), but with the other improvements it is only about an extra minute to total build time. Builds are ordered for optimal running time when Travis is running normally, but still getting important results quickly if overloaded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,20 @@
+language: cpp
+sudo: false
+dist: trusty
 addons:
   apt:
     packages:
+      - g++-4.8
+      - g++-4.8-multilib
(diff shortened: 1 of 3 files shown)


asherkin:
That is why I reordered them - the 6th job starts and finishes before the slowest job of the initial 5 completes.

dvander:
Nice!

PR #180, opened 2018-02-14
Added __LINE__ const

CrazyHackGUT:
Added `__line` const. He always equals current code line.

Test code:
```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    PrintToServer("Check magic Pawn consts...");

    PrintToServer("cellbits = %d", cellbits);
    PrintToServer("cellmax  = %d", cellmax);
    PrintToServer("cellmin  = %d", cellmin);

    PrintToServer("charbits = %d", charbits);
    PrintToServer("charmax  = %d", charmax);
    PrintToServer("charmin  = %d", charmin);

    PrintToServer("debug    = %d", debug);

    PrintToServer("__Pawn   = %x", __Pawn);
    PrintToServer("__line   = %d", __line);

    PrintToServer("ucharmax = %d", ucharmax);
}
```

example running test code
line number


Drifter321:
This should probably be __LINE__ to be the same as the amxmodx one since it matches other languages. https://github.com/alliedmodders/amxmodx/blob/1a56577f7a3c8167f34cec0bbd443402f5636dcf/compiler/libpc300/sc1.c#L1463

compiler/sc2.cpp
```diff
@@ -388,6 +389,9 @@ static void readline(unsigned char *line)
       line+=strlen((char*)line);
     } /* if */
     fline+=1;
+    sym=findconst("__line",NULL);
+    assert(sym!=NULL);
+    sym->addr_=fline;
```
rsKliPPy: You should probably do `sym->setAddr(fline);` instead to be consistent.

CrazyHackGUT: Thanks.

dvander:
this looks fine but it deserves additions to compiler/tests - ideally two, one testing that __LINE__ works as normal and another that it works in macro expansions.

CrazyHackGUT:
Thanks. I added tests.

dvander:
Thanks!

PR #184, opened 2018-03-05
Optimize compilation for large global arrays

rsKliPPy:
Introduces `dumpdup` opcode (compiler only). It helps with huge arrays with default values by creating only 1 `dumpdup` opcode instead of many `dump` opcodes together with a lot of vaues. Array indirection tables still use `dump` though, but they are small in size compared to the array data itself.

The following array:
```SourcePawn
bool x[32][1024][1024];
```
compiles in ~1.7s with this patch instead of ~9.5s on my machine.

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -1363,7 +1363,7 @@ static void dumpzero(int count)
     return;
   assert(curseg==2);
   
-  stgwrite("dumpdup ");
+  stgwrite("dumpfill ");
   outval(0, FALSE);
   stgwrite(" ");
   outval(count, TRUE);
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -1359,20 +1359,14 @@ static void dumplits(void)
  */
 static void dumpzero(int count)
 {
-  int i;
-
   if (sc_status==statSKIP || count<=0)
     return;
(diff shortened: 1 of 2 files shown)


compiler/sc1.cpp
```diff
@@ -1359,20 +1359,14 @@ static void dumplits(void)
...
-  i=0;
-  while (count-- > 0) {
-    outval(0, FALSE);
-    i=(i+1) % 16;
-    stgwrite((i==0 || count==0) ? "\n" : " ");
-    if (i==0 && count>0)
-      defstorage();
-  } /* while */
+  
+  stgwrite("dumpdup ");
```
dvander: This change seems good - could you call it "dumpfill" or something instead? "dup" usually means something else in VMs and there's precedent for "fill" here.

rsKliPPy: Sure, I wasn't keen on the naming anyway. I chose that naming because of ASM `DUP(value)`.
Do you want me to amend or create a new commit?

dvander: Up to you - I'll squash it in the end.

rsKliPPy: Out of curiosity - what's the usual meaning of "dup" in VMs?

dvander:
Thanks!

PR #190, opened 2018-03-24
Fix register increment/decrement visit

Kailo97:
Probably copy-paste mistake.

### DIFF
--- a/vm/interpreter.cpp
+++ b/vm/interpreter.cpp
@@ -393,7 +393,7 @@ Interpreter::visitINVERT()
 bool
 Interpreter::visitINC(PawnReg dest)
 {
-  regs_.pri() += 1;
+  regs_[dest] += 1;
   return true;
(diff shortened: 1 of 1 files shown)


dvander:
Looks good, nice catch - was there a test case that caused this, that we could add to the suite?

Kailo97:
Just found this while looking for code (writing bytecode reference). sc4.cpp saying that "inc/dec.alt" not generatable by original SP compiler. But it can be generated by "emit" supported compiler or as direct smx modification.

dvander:
Hrm, okay. With ungenerated opcodes I'd prefer to remove them from the verifier & VM. It makes the implementation surface a bit smaller. Lemme first make sure it didn't appear in older versions or anything though.

dvander:
Well, it's never been used, but might as well keep it around. It's already implemented.

PR #193, opened 2018-04-12
Remove #pragma rational.

dvander:
There's a snowball's chance in hell that we're going to treat rational literals as fixed-point, so I think it's safe to remove `#pragma rational`.


Fyren:
This is _irrational_.

dvander:
You’re grounded!

PR #206, opened 2018-05-03
Add look-ahead for oldstyle decl

### LINKED ISSUE #162: Anonymous Instantiation Failure
Anonymous methodmap instantiation fails because spcomp expects an olddecl int declaration, but it's not.

```cpp
/*
plugin.sp(14) : warning 219: local variable "ArrayList" shadows a variable at a preceding level
plugin.sp(14) : error 001: expected token: ";", but found "("
plugin.sp(14) : error 029: invalid expression, assumed zero
plugin.sp(14) : error 001: expected token: ")", but found ";"
plugin.sp(14) : fatal error 190: too many error messages on one line
*/
#include <sourcemod>

public void OnPluginStart()
{
	new ArrayList();
}
```

Headline:
This commit prevents anonymous instantiations in the form of `new Class()` from falling into declloc.

Fixes #162

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -5391,10 +5391,10 @@ static void statement(int *lastindent,int allow_decl)
   case tDECL:
   case tSTATIC:
   case tNEW:
-    if (matchtoken(tSYMBOL)) {
+    if (tok == tNEW && matchtoken(tSYMBOL)) {
         if (lexpeek('(')) {
           lexpush();
-          goto doxpr_jmp;
+          goto doexpr_jump;
         }
       lexpush(); // we matchtoken'ed, give it back to lex for declloc
     }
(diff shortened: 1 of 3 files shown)


Headline:
It seems this fails many tests, I'll investigate in the near future.

Headline:
Cool. Tests are passing 🍻

compiler/sc1.cpp
```diff
@@ -5391,6 +5391,13 @@ static void statement(int *lastindent,int allow_decl)
   case tDECL:
   case tSTATIC:
   case tNEW:
+    if (matchtoken(tSYMBOL)) {
```
dvander: I think you want this to be `if (tok == tNEW && matchtoken(tSYMBOL)) {` otherwise something like `static MyMethodMap()` would go into doexpr.

compiler/sc1.cpp
```diff
@@ -5391,6 +5391,13 @@ static void statement(int *lastindent,int allow_decl)
   case tDECL:
   case tSTATIC:
   case tNEW:
+    if (matchtoken(tSYMBOL)) {
+        if (lexpeek('(')) {
+          lexpush();
+          goto doxpr_jmp;
```
dvander: nit: s/doxpr_jmp/doexpr_jmp/

dvander:
Thanks! Looks good.

PR #207, opened 2018-05-04
Fix a HEA leak when using break in nested scopes.

dvander:
Ancient bug - looks like it's a carryover from AMX Mod X.

The following script will misalign the HEA stack:

```
native void printnum(int n);

public main() {
  while (true) {
    for (int i = 1; i < 10; i++) {
      char[] egg = new char[i];
      if (i % 2 == 0) {
         char[] yam = new char[i + 3];
         break;
      } else {
         char[] yam = new char[i + 3];
         continue;
      }
    }
  }
}
```

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -6244,6 +6244,7 @@ static void dobreak(void)
     return;
   destructsymbols(&loctab,nestlevel);
   genstackfree(ptr[wqBRK]);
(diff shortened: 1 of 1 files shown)


dvander:
Not from AMX Mod X, but was improperly fixed 12 years ago: https://github.com/alliedmodders/sourcemod/commit/4d44a1df4fd60db66e7f4b793ad29ef17a9fec1a

PR #208, opened 2018-05-06
Remove pow10 usage and shim.

psychonic:
Remove usage of `pow10`. This is documented as nonstandard GNU extension.

We already needed to have a shim for it on macOS, and using newer versions of Clang (tested v6) on Linux also fail to find it, even after explicitly including <math.h>. (I haven't tried newer GCC versions).

### DIFF
--- a/compiler/sc2.cpp
+++ b/compiler/sc2.cpp
@@ -65,13 +65,6 @@ static short skiplevel; /* level at which we started skipping (including nested 
 static unsigned char term_expr[] = "";
 static int listline=-1; /* "current line" for the list file */
 
-#if defined KE_MACOSX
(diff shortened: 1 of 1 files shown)


Amaroq7:
I'm also unable to compile the compiler due that function. I've tried to compile using both Clang 6 and GCC 8.1. It seems it's been removed since the 2.27 glibc release.

```
* The libm functions pow10, pow10f and pow10l are no longer supported for
  new programs.  Programs should use the standard names exp10, exp10f and
  exp10l for these functions instead.
```
Source: https://lists.gnu.org/archive/html/info-gnu/2018-02/msg00000.html

PR #209, opened 2018-05-09
Refactor some common code into a shared library.

dvander:


dvander:
The purpose of this is twofold: one, introduce a "libsmx" helper library to handle basic SMX parsing and creating. Second, a bunch of stuff is awkwardly duplicated between exp/compiler and compiler, and I'd like to make it common.

PR #210, opened 2018-05-14
Add RTTI metadata.

dvander:
This is a major overhaul of the SMX layout, focused on giving more information to the VM, compacting existing information, and reducing the number of difficult, variable-length tables to decode.

First, new tables have been introduced to fully describe all SourcePawn types (except for, of course, enum structs). These tables are `rtti.enums`, `rtti.classdefs`, `rtti.fields`, `rtti.natives`, `rtti.typedefs`, `rtti.typesets`, and `rtti.methods`. These tables will be required as new features depend on them. They are essential for future improvements to the language, such as direct plugin linkage, fully dynamic arrays, struct support, and removing the native thunk.

All of these tables are fixed-length. In a few cases, we do need to encode variable length data. For example, the types of struct fields, the dimensions of arrays, the signatures of methods, etc. This information has a compact byte encoding specified in `smx-typeinfo.h`. The encoding is stored in a blob section called `rtti.data`. In some cases, the encoding can fit directly in a 32-bit integer field, and the blob table does not have to be used. When it is used, it performs basic compression by deduplicating entries.

In addition to these new tables, the old debug tables have been overhauled as well:

 - `dbg.natives` has been replaced by `rtti.natives`.
 - `dbg.symbols` has been removed.
 - `dbg.strings` has been removed.
 - `dbg.tags` has been removed.
 - `dbg.locals` and `dbg.methods` now contain local variable information.
 - `dbg.globals` now contains global variable information.

Since the new debug tables use RTTI, it is heavily compacted compared to the older tables. For example, funcommands.smx went from 34K to 29K (a 15% reduction).

All rtti tables are implicitly versioned. Each has a header_size field that allows for growing the header, and each has a row_size field in case the fixed-length data needs to change.

Also, smxviewer and smxdasm have been updated to support debugging these new tables.

I would like to pin SourceMod's development branch to these changes soon.


dvander:
There's no native int in the language - but if there was, I guess my question would be what are the use cases beyond storing pointers? Marshalling things like `ssize_t` might be cleaner, maybe?

assyrianic:
possible use case beyond pointers is having platform-independent sizing similar to `size_t` but that's the only other use case. Either using to store pointers or platform independent sizing.

PR #211, opened 2018-05-16
Fix AppVeyor and tests.

Fyren:
P-M pointed out it seemed like AppVeyor wasn't doing... much of anything.  Upon investigation, AppVeyor wasn't doing much of anything.  Also turns out the newer test script stuff was broken on Windows anyway and was only checking for the first error line of failure tests (and was even doing _that_ improperly.)


Fyren:
I think it broke in early October.

PR #216, opened 2018-05-27
Reenable callback on |break| instruction

peace-maker:
~~Reenable the `IPluginContext::SetDebugBreak`~~ Add `ISourcePawnEngine::SetDebugBreakHandler` to allow others to listen for every `break` instruction in the code stream. Add functions to `IPluginDebugInfo` to allow finding the address of functions and lines in plugins for break points.

This doesn't expose any local nor global variable information yet, so debugging capabilities are pretty limited. You can print which file and line the plugin is about to execute currently.

This is partly extracted from #51.

### DIFF
--- a/include/sp_vm_types.h
+++ b/include/sp_vm_types.h
@@ -111,7 +111,7 @@ namespace SourcePawn
 	class IPluginContext;
 	class IVirtualMachine;
 	class IProfiler;
-  class IErrorReport;
+	class IErrorReport;
 };
 
 struct sp_context_s;
--- a/include/sp_vm_types.h
+++ b/include/sp_vm_types.h
@@ -55,6 +55,8 @@ typedef uint32_t	funcid_t;			/**< Function index code */
 #define SP_PROF_CALLBACKS		(1<<1)		/**< Profile callbacks. */
 #define SP_PROF_FUNCTIONS		(1<<2)		/**< Profile functions. */
 
+#define DEBUG_BREAK_INFO_VERSION 0x0001 /**< Version of the sp_debug_break_info_t struct. */
+
 /**
  * @brief Error codes for SourcePawn routines.
  */
@@ -248,14 +250,28 @@ typedef struct sp_debug_symbol_s
 } sp_debug_symbol_t;
 
 /**
- * Breaks into a debugger
+ * @brief Context describing the VM state when the SPVM_DEBUGBREAK
+ * callback is called.
+ */
+typedef struct sp_debug_break_info_s
+{
+  uint16_t  version;   /**< Version of this struct */
+  cell_t    cip;       /**< Current virtual instruction pointer */
+  cell_t    frm;       /**< Current virtual frame pointer */
+  cell_t    sp;        /**< Current virtual stack pointer */
+  uint8_t*  memory;    /**< Pointer to raw plugin image in memory */
+  size_t    mem_size;  /**< Total size of the plugin image in memory */
+} sp_debug_break_info_t;
+
+/**
+ * Breaks into a debugger.
+ * If the exception parameter is not null, 
+ * the function is called with the location of the error.
  * Params:
  *  [0] - plugin context
- *  [1] - frm
- *  [2] - cip
- *  [3] - memory
- *  [4] - exception
+ *  [1] - debug info
+ *  [2] - exception
  */
-typedef void(*SPVM_DEBUGBREAK)(SourcePawn::IPluginContext *, cell_t, cell_t, void *, const SourcePawn::IErrorReport *);
+typedef void(*SPVM_DEBUGBREAK)(SourcePawn::IPluginContext*, sp_debug_break_info_t&, const SourcePawn::IErrorReport*);
 
 #endif //_INCLUDE_SOURCEPAWN_VM_TYPES_H
--- a/include/sp_vm_types.h
+++ b/include/sp_vm_types.h
@@ -258,9 +258,6 @@ typedef struct sp_debug_break_info_s
   uint16_t  version;   /**< Version of this struct */
   cell_t    cip;       /**< Current virtual instruction pointer */
   cell_t    frm;       /**< Current virtual frame pointer */
-  cell_t    sp;        /**< Current virtual stack pointer */
-  uint8_t*  memory;    /**< Pointer to raw plugin image in memory */
-  size_t    mem_size;  /**< Total size of the plugin image in memory */
 } sp_debug_break_info_t;
 
 /**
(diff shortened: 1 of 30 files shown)


vm/smx-v1-image.cpp
```diff
@@ -756,3 +756,153 @@ SmxV1Image::LookupLine(uint32_t addr, uint32_t* line)
...
+    cursor += sizeof(SymbolType);
+  }
+  return false;
+}
+
+bool
+SmxV1Image::LookupFunctionAddress(const char* function, const char* file, ucell_t* funcaddr)
+{
+  uint32_t index = 0;
+  const char* tgtfile;
```
dvander: nit: move tgtfile down to where it's actually used. Same for index. (We're not using C :p)

vm/x86/jit_x86.cpp
```diff
@@ -1116,6 +1117,14 @@ Compiler::visitREBASE(cell_t addr, cell_t iv_size, cell_t data_size)
   return true;
 }
 
+bool
+Compiler::visitBREAK()
+{
+  __ call(&debug_break_);
```
dvander: As discussed this should not be emitted unless the debugger is attached. It will definitely have a performance cost.

vm/x86/jit_x86.cpp
```diff
@@ -1601,6 +1610,31 @@ Compiler::emitThrowPath(int err)
...
+  __ subl(tmp, dat);
+  __ movl(Operand(spAddr()), tmp);
+
+  // Enter the exit frame. This aligns the stack.
+  __ enterExitFrame(ExitFrameType::Helper, 0);
+
+  // Get the context pointer and call the debugging break handler.
+  __ push(0); // IErrorReport*
+  __ push(intptr_t(rt_->GetBaseContext()));
+  __ call(ExternalAddress((void *)InvokeDebugger));
```
dvander: I believe you need to keep the stack aligned here - it looks like it'll be 8 bytes at the call, when it should be 16. (This matters because GCC assumes an aligned stack and certain SSE instructions will fault if they have an unaligned address.)

peace-maker: `enterExitFrame` pushes 3 parameters as well though. I'm confused. Should I still just add
```cpp
__ subl(esp, 8);
```
after calling `enterExitFrame` and do
```cpp
__ addl(esp, 16);
```
after the call?

https://github.com/alliedmodders/sourcepawn/blob/cf5020e8e825fb3647d6e578ff372deb563240c6/vm/x86/macro-assembler-x86.h#L46-L61

I've copied that comment from `emitCallThunk`
https://github.com/alliedmodders/sourcepawn/blob/cf5020e8e825fb3647d6e578ff372deb563240c6/vm/x86/jit_x86.cpp#L1244-L1245

include/sp_vm_types.h
```diff
@@ -246,4 +249,26 @@ typedef struct sp_debug_symbol_s
 	sp_debug_symbol_raw_t *sym;	 /**< Pointer to original symbol */
 } sp_debug_symbol_t;
 
+/**
+ * @brief Context describing the VM state when the SPVM_DEBUGBREAK
+ * callback is called.
+ */
+typedef struct sp_debug_break_info_s
+{
+  uint16_t  version;   /**< Version of this struct */
```
dvander: How is this version field supposed to be used?

vm/debugging.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "environment.h"
+#include "watchdog_timer.h"
+#include <amtl/am-raii.h>
+
+namespace sp {
+
+void InvokeDebugger(PluginContext* ctx, const IErrorReport* report)
+{
+  // Ignore any calls if this isn't enabled.
+  if (!Environment::get()->IsDebugBreakEnabled())
```
dvander: Is it possible to get here if the debug break is not enabled?

vm/smx-v1-image.cpp
```diff
@@ -756,3 +756,153 @@ SmxV1Image::LookupLine(uint32_t addr, uint32_t* line)
...
+        break;
+      index++;
+      assert(index < debug_info_->num_syms);
+    }
+  }
+
+  // now find the first line in the function where we can "break" on
+  uint32_t index = 0;
+  for (; index < debug_info_->num_lines && debug_lines_[index].addr < *funcaddr; index++)
+    /* nothing */;
```
dvander: nit: prefer {} for empty statements, or `continue`

vm/smx-v1-image.cpp
```diff
@@ -756,3 +756,153 @@ SmxV1Image::LookupLine(uint32_t addr, uint32_t* line)
...
+  */
+  *addr = 0;
+
+  uint32_t bottomaddr, topaddr;
+  uint32_t file;
+  uint32_t index = 0;
+  for (file = 0; file < debug_info_->num_files; file++) {
+    // find the (next) matching instance of the file
+    if (debug_files_[file].name >= debug_names_section_->size ||
+      strcmp(debug_names_ + debug_files_[file].name, filename) != 0)
```
dvander: nit: indent `strcmp` to line up with `debug_files_`

vm/smx-v1-image.cpp
```diff
@@ -756,3 +756,153 @@ SmxV1Image::LookupLine(uint32_t addr, uint32_t* line)
...
+
+  *funcaddr = debug_lines_[index].addr;
+  return true;
+}
+
+bool
+SmxV1Image::LookupLineAddress(const uint32_t line, const char* filename, uint32_t* addr)
+{
+  /* Find a suitable "breakpoint address" close to the indicated line (and in
+  * the specified file). The address is moved up to the next "breakable" line
```
dvander: nit: Line up `*` or use C++-style comments

vm/smx-v1-image.cpp
```diff
@@ -756,3 +756,153 @@ SmxV1Image::LookupLine(uint32_t addr, uint32_t* line)
...
+    bottomaddr = debug_files_[file].addr;
+    topaddr = (file + 1 < debug_info_->num_files) ? debug_files_[file + 1].addr : (uint32_t)-1;
+
+    // go to the starting address in the line table
+    while (index < debug_info_->num_lines && debug_lines_[index].addr < bottomaddr)
+      index++;
+
+    // browse until the line is found or until the top address is exceeded
+    while (index < debug_info_->num_lines &&
+      debug_lines_[index].line < line &&
```
dvander: nit: Line up `debug_lines_` with `index`

vm/debugging.cpp
peace-maker: Yes, through the interpreter.

include/sp_vm_types.h
peace-maker: To know whether new fields are available. When adding more info to the struct, like maybe a pointer to the smx in memory to parse the debug symbols manually, allow extensions know which fields they can use.

Probably could get rid of this seperate version field and just increase the whole vm api version if it's changed.

PR #217, opened 2018-05-28
JIT off the control-flow graph, rather than instruction stream.

dvander:
It is much simpler to generate bytecode off basic blocks, instead of the raw instruction stream. We don't need to store the gross and memory-intensive jump map. Instead, the address of every target is contained in the target block itself.


assyrianic:
sounds like a good idea, what are the design pros and cons of JITing the graph blocks as opposed to the regular instruction stream?

dvander:
The old way the JIT worked was to keep a hashtable with the address of every instruction, so we could find all the locations of jump targets. It's super gross and there was never any rigorous error checking that the hash table was filled (i.e., that jump targets were valid).

Working directly off the CFG eliminates that whole mess. The target of a jump or switch is just a block now. No need to cache and lookup addresses.

PR #238, opened 2018-09-28
Annihilate codepage & resolve startup segfault 

Headline:
I think this might have been the same case @KyleSanderson ran into. Not sure of it's cause, or why it has recently become a problem, but this should handle it just fine.


compiler/sc1.cpp
```diff
@@ -1088,14 +1088,16 @@ static void setconfig(char *root)
 
       /* same for the codepage root */
 # if !defined NO_CODEPAGE
-        *ptr='\0';
+        if(ptr)
```
dvander: I don't see a way ptr can be NULL here, as it's NULL checked above:

`  if ((ptr=strrchr(path,DIRSEP_CHAR))!=NULL || (ptr=strchr(path,':'))!=NULL) {`

Headline: i think it’s reassigned and catches the plague here
https://github.com/alliedmodders/sourcepawn/blob/425c86a8f9d40c8bffec7b4bd2e6d32233c66f8f/compiler/sc1.cpp#L1077

Headline: for reference, here's my trace from the segfault
screen shot 2018-09-28 at 7 16 52 pm

dvander:
Thanks!

PR #239, opened 2018-09-28
SmxViewer: Collect unknown functions while disassembling

peace-maker:
When encountering a `call` while disassembling check if there is an entry for the called address in the ".publics" or ".dbg.symbols" sections. If the function is missing add it to a seperate list to be able to disassemble functions even while debug info is stripped.

The functions are added with a name like `sub_<addr>` to be unqiue.

### DIFF
--- a/tools/smxtools/smxdasm/File.cs
+++ b/tools/smxtools/smxdasm/File.cs
@@ -1,5 +1,6 @@
 ﻿using System.IO;
 using System.Collections.Generic;
+using System;
 
 namespace smxdasm
 {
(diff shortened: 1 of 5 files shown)


tools/smxtools/smxdasm/File.cs
```diff
@@ -131,6 +135,52 @@ public SmxFile(BinaryReader br)
...
             UnknownSections = unknown.ToArray();
+
+            // Disassemble all functions right away to find all called functions.
+            if (DebugSymbols != null)
+            {
+                foreach (var entry in DebugSymbols.Entries)
+                {
+                    if (entry.Ident != SymKind.Function)
+                        continue;
+                    try
```
dvander: Rather than duplicate this try block 3 times, could you have a helper function like "TryDisassemble"?

PR #244, opened 2018-10-01
Remove some stuff from symbol.

dvander:
This is not used in SourceMod, and SMX does not have a library table.


Fyren:
We don't accept commit messages with emojis here, dvander.

dvander:
That must be why travis died.

PR #249, opened 2018-10-03
More sctracker cleanups.

dvander:


compiler/sctracker.cpp
```diff
@@ -384,15 +383,22 @@ stack_scope_id()
...
+methodmap_t::methodmap_t(methodmap_t* parent, LayoutSpec spec, const char* name)
+ : parent(parent),
+   tag(0),
+   nullable(false),
+   keyword_nullable(false),
+   spec(spec),
+   dtor(nullptr),
+   ctor(nullptr)
+{
+  ke::SafeStrcpy(this->name, sizeof(this->name), name);
```
Headline: any reason to keep `name` in `methodmap_t` a static sized char array as opposed to just a `ke::AString`? Then you could just use the constructor in the initializer list while removing a bad `sNAMEMAX`

compiler/sc1.cpp
```diff
@@ -3451,21 +3452,21 @@ methodmap_method_t *parse_method(methodmap_t *map)
...
-    return NULL;
+    return nullptr;
 
-  methodmap_method_t *method = (methodmap_method_t *)calloc(1, sizeof(methodmap_method_t));
+  auto method = ke::MakeUnique<methodmap_method_t>(map);
   strcpy(method->name, ident.name);
   method->target = target;
-  method->getter = NULL;
-  method->setter = NULL;
+  method->getter = nullptr;
```
Headline: Are these needed any longer since you already init to `nullptr` in the `methodmap_method_t` constrcutor

compiler/sctracker.cpp
```diff
@@ -431,42 +431,18 @@ methodmap_t *methodmap_find_by_name(const char *name)
...
-    methodmap_t *next = ptr->next;
-    for (size_t i = 0; i < ptr->nummethods; i++)
-      free(ptr->methods[i]);
-    free(ptr->methods);
-    free(ptr);
-    ptr = next;
-  }
-  methodmap_first = NULL;
-  methodmap_last = NULL;
+  sMethodmaps.clear();
```
Headline: Nice!

compiler/sc1.cpp
dvander: Nope, thanks, I'll drop these lines.

compiler/sctracker.cpp
dvander: I dislike AString enough that I'm currently debating whether to go whole-hog and move spcomp to std::string (and later vector). Since it's a standalone executable there should be no problems.

But, it's a change in semantics, since sNAMEMAX is embedded everywhere. I'd want to open that can of worms in a separate patch.

Headline: That makes sense, and adding std::string is probably a good move as well. 👍

PR #256, opened 2018-10-17
Fix assert when using a methodmap before declaring it.

dvander:
This partially fixes alliedmodders/sourcemod#897.

### DIFF
--- a/compiler/sc5-in.scp
+++ b/compiler/sc5-in.scp
@@ -35,7 +35,7 @@ static const char *errmsg[] = {
 /*008*/  "must be a constant expression; assumed zero\n",
 /*009*/  "invalid array size (negative, zero or out of bounds)\n",
 /*010*/  "invalid function or declaration\n",
-/*011*/  "unused11\n",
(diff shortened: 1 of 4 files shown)


compiler/sc5-in.scp
```diff
@@ -35,7 +35,7 @@ static const char *errmsg[] = {
 /*008*/  "must be a constant expression; assumed zero\n",
 /*009*/  "invalid array size (negative, zero or out of bounds)\n",
 /*010*/  "invalid function or declaration\n",
-/*011*/  "unused11\n",
+/*011*/  "symbol '%s' already defined - did you use a methodmap before its declaration?\n",
```
KyleSanderson: I like this message but the compiler usually doesn't ask questions of its users. If we're `HAL 9000`ing this I absolutely love it.

dvander: I think it's okay. I swear I've seen clang do it. I'm hoping this message will be temporary though, as I'd like to fix the underlying bug.

KyleSanderson: All good bud ;)

PR #262, opened 2018-11-06
Minor performance enhancements to spcomp.

dvander:


dvander:
We could probably flip STL on in spcomp any time, since it's standalone.

PR #265, opened 2018-11-10
Fix building GCC 8

Amaroq7:
Fixes building using gcc 8. Currently it fails due ```-Werror=class-memaccess```

> -Wclass-memaccess warns when objects of non-trivial class types are manipulated in potentially unsafe ways by raw memory functions such as ```memcpy```, or ```realloc```.
> The warning helps detect calls that bypass user-defined constructors or copy-assignment operators, corrupt virtual table pointers, data members of const-qualified types or references, or member pointers.
> The warning also detects calls that would bypass access controls to data members.

[Source](https://gcc.gnu.org/gcc-8/changes.html)

### DIFF
--- a/compiler/sc1.cpp
+++ b/compiler/sc1.cpp
@@ -3797,7 +3797,7 @@ static void dodelete()
  *                 | function-type-inner
  * function-type-inner ::= "function" type-expr "(" new-style-args ")"
  */
-static void parse_function_type(ke::UniquePtr<functag_t> &type)
+static void parse_function_type(const ke::UniquePtr<functag_t> &type)
 {
   int lparen = matchtoken('(');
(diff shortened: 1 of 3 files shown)


compiler/sc1.cpp
```diff
@@ -3798,10 +3797,8 @@ static void dodelete()
  *                 | function-type-inner
  * function-type-inner ::= "function" type-expr "(" new-style-args ")"
  */
-static void parse_function_type(functag_t *type)
+static void parse_function_type(ke::UniquePtr<functag_t> &type)
```
dvander: It looks like this signature didn't need to change? If so it should take a `functag_t*` or a `const ke::UniquePtr<functag_t>&`.

PR #267, opened 2018-11-18
It's been, like, 20 years. Rename files.

dvander:
Since Pawn was "SmallC" it's used a numbering scheme for its source
files, roughly numbered after stages of compilation:

 1 - Parser entry point, statement parsing.
 2 - Lexing.
 3 - Expression parsing.
 4 - Code generation.
 5 - Error reporting.
 6 - Assembly.
 7 - Peephole optimization.

While Pawn "veterans" know this numbering by now, it has absolutely no
meaning to anyone else. So let's call these files what they are.


compiler/errors.cpp
```diff
@@ -37,14 +37,14 @@
...
-#include "sc5.h"
+#include "errors.h"
 
 #if defined _MSC_VER
   #pragma warning(push)
   #pragma warning(disable:4125)  /* decimal digit terminates octal escape sequence */
 #endif
 
-#include "sc5-in.scp"
+#include "messages.scp"
```
Headline: while we're doing renaming, is there any reason .scp is preferred over .cpp?

dvander: Probably a better name would be ".tbl" or something. It's not really a typical source file or header. I've no idea what "scp" meant.

dvander: Oh, these things actually *are* headers... guess they should be .h then.

PR #268, opened 2018-11-18
Split sc.h into multiple headers.

dvander:
I don't know why every single thing is shoved into sc.h. This patch reduces the clutter a bit.


compiler/assembler.h
```diff
@@ -0,0 +1,25 @@
+// vim: set ts=8 sts=2 sw=2 tw=99 et:
+//
+//  Pawn compiler - Error message system
```
peace-maker: Every new header file has this comment about an `Error message system`. Might want to remove that / change it.

compiler/parser.cpp
```diff
@@ -5428,10 +5436,6 @@ static void statement(int *lastindent,int allow_decl)
     doassert();
     lastst=tASSERT;
     break;
-  case tSLEEP:
```
peace-maker: Did you mean to slip this removal into this commit?

dvander: Yes, indeed.

compiler/assembler.h
dvander: Thanks, I'll fix this up in the next commit.

PR #271, opened 2018-11-19
Remove SC_LIGHT defines.

dvander:


compiler/lexer.cpp
```diff
@@ -489,7 +481,6 @@ static void stripcom(unsigned char *line)
       } else if (*line=='/' && *(line+1)=='/'){  /* comment to end of line */
         if (strchr((char*)line,'\a')!=NULL)
           error(49);    /* invalid line continuation */
-        #if !defined SC_LIGHT
           if (*(line+2)=='/' && *(line+3)<=' ') {
```
Fyren: Missed the unindent here.

dvander: Thanks, I got confused by the unbraced if above.

PR #282, opened 2018-12-02
Replace the macro list with a hash table.

dvander:
This removes the last use of stringpair and along with it more manual
memory management and C-era linked list code. There is no performance
difference because, amazingly, there was a tiny open-coded lookup cache
in the macro code. Go figure.


dvander:
Heh, good catch. delete_subst was returning TRUE/FALSE which I preserved, but command() does not return boolean. I have no idea why it was doing this and it was probably incorrect. Fixed now.

PR #287, opened 2018-12-04
Add transitional syntax for enum structs.

dvander:
Add transitional syntax for enum structs.

This amends the Transitional Syntax to allow enum structs, codifying a
permanent replacement and obsoleting the old syntax. The grammar is:

    enum struct <name> { <newline>
      [<new-type-expr> <name> [post-dims] <newline>]*
    } <newline>

Enum structs declared this way can be used anywhere a new-style type is
declared. Internally, they still decompose into arrays. As such,
anywhere an array is forbidden, enum structs are not allowed.
Similarly, if used in an array, an enum struct type will become a
multi-dimensional array.

Unlike old-style enum structs, they are not accessed with array syntax.
Instead they are accessed through the field operator ("."). For example
an old enum struct might look like:

    enum Sample {
      Float:x, Float:y
    };
    new s[Sample];
    s[x] = 10; s[y] = 20;

A new-style example might look like:

    enum struct Sample {
      float x;
      float y;
    };
    Sample s;
    s.x = 10; s.y = 20;

Similarly, the field names of a new-style enum struct do not poison the
global scope. In fact they are not accessible, and the enum struct name
itself does not define a global symbol (only a named type).

Enum structs can also have inline methods, similar to methodmaps.
Unlike methodmaps, these methods cannot be native, and do not have
public or private decorators. Like methodmaps, they must use new-style
syntax.

This new syntax obsoletes old-style enum structs. They will no longer be
maintained going forward, and bugs in them will not be fixed. They may
be broken in the future.

Rationale:

Enum structs have been neglected due to having a painfully complex and
strange implementation. Its problems roughly fall into three categories:
 1. Syntax: they are awkwardly layered on top of enums and arrays.
 2. Semantics: they define unscoped globals that expose implementation
    details, and can be used in illegal, nonsensical ways.
 3. Implementation: rather than define a type, they create a shadow
    structure attached to arrays, which then requires complex pattern
    matching to reconstruct.

Developers have been asking for struct syntax, even if it's sugar around
enum structs. In the past we've been reluctant to do this, because
Structs remain a large carrot for moving to a newer compiler. Properly
implemented, they allow many things enum structs do not (including
cross-plugin and native support). Syntactic sugar does not fix the deeper
problems in the runtime, and to fix those, we would need users to accept
backwards compatibility breaks and experimental behavior. Thus, structs
would be the carrot to entice this.

However this thinking was probably mistaken. No "correct" struct
implementation has been forthcoming, and meanwhile, we're stuck with a
poorly implemented and widely used language misfeature. It is better to
provide something maintainable to promote better plugins, and provide
developers with a better environment. This solution is also incomplete:
there are many things enum structs still don't allow, so we have not
lost a carrot for future versions.


dvander:
TODO: finish RTTI and update smx viewer.

Headline:
can we drop the `enum` from `enum struct`?

If syntax is identical to how one would expect typical struct usage, why should we treat it differently?

dvander:
The biggest reason is that I'm not positive it's identical. I haven't thought about it carefully enough, but it seems like there could be differences given that it's still sugar for arrays. Like, pass-by-value vs pass-by-reference.

Another reason is that we already have a "struct" keyword (see core.inc). They have different semantics. The pre-existing keyword allows things like `char[]` in fields which enum structs do not. Unifying them would be tricky, but not impossible. Basically the pre-existing cases have to fall into declstruct() and the new cases would have to fall into initials(). We could get there with pattern matching, and probably delete the old pstruct code.

The potential for semantic difference is a bigger issue, but I'm nowhere near 100% sold on "enum struct" syntax or future-structs working one way or another.

Headline:
The only reason I suggested that is because I can't see a difference in how this behaves vs real structs (despite the future feature set that real structs would get). 

Since in sourcepawn arrays are always pass-by-reference, that seems inline with how a struct would behave, unless you had a different idea around it. I think it'd be a *little* misleading to some people, but 90% of the consumers wouldn't be able to identify whether these structs are "real" or not, which in my opinion is a good thing.

TotallyMehis:
> Internally, they still decompose into arrays

Are we able to get the position of a member and the size of the struct somehow?


Right now you do things like this if you want a global dynamic array with a pseudo struct:
```sourcepawn
enum
{
    FOO_X = 0,
    FOO_Y,
    FOO_Z_FLOAT,
    
    FOO_SIZE
};

ArrayList globalarray;

public void OnPluginStart()
{
    globalarray = new ArrayList( FOO_SIZE );
    
    for ( int i = 0; i < 10; i++ )
    {
        any data[FOO_SIZE];
        data[FOO_X] = 1;
        data[FOO_Y] = 2;
        data[FOO_Z_FLOAT] = view_as<int>( 3.0 );
        
        globalarray.PushArray( data );
    }
    
    
    float z = view_as<float>( globalarray.Get( 0, FOO_Z_FLOAT ) );
    PrintToServer( "z_float: %.1f", z );
}
```

Something like this would be great, imo:
```sourcepawn
enum struct Foo
{
    int x;
    int y;
    float z;
};

ArrayList globalarray;

public void OnPluginStart()
{
    globalarray = new ArrayList( sizeof( Foo ) ); // ???
    
    for ( int i = 0; i < 10; i++ )
    {
        Foo f;
        f.x = 1;
        f.y = 2;
        f.z = 3.0;
        
        globalarray.PushArray( view_as<any>( f ) ); // ???
    }
    
    
    float z = view_as<float>( globalarray.Get( 0, Foo::z ) ); // ???
    PrintToServer( "z_float: %.1f", z );
}
```

assyrianic:
I have created a tuple type that acts exactly like a C/C++ struct, including packing and padding features. would it be possible to back the enum struct with the tuple as a handle?

dvander:
> Since in sourcepawn arrays are always pass-by-reference, that seems inline with how a struct would behave, unless you had a different idea around it.

In almost every other language, arrays are pass-by-ref and structs are pass-by-value. I don't know if that's more or less confusing or not. But probably it'd be confusing if structs and enum-structs behaved differently.

dvander:
> > Internally, they still decompose into arrays
> 
> Are we able to get the position of a member and the size of the struct somehow?

I'll give some thought about how to integrate this with native data structures. Exposing sizeof() and offsets seems like a mistake to me. If we wind up dropping the "enum" from this syntax as Headline suggested, we absolutely don't want that. The fact that it is an array internally must be completely hidden. Otherwise, adding proper struct features in the future will get much more difficult, since we'll have to make sure array-like hacks continue to work.

If we keep the "enum struct" syntax, then it might be okay, since it's clearly a separate feature.

dvander:
> I have created a tuple type that acts exactly like a C/C++ struct, including packing and padding features. would it be possible to back the enum struct with the tuple as a handle?

This is intended to be a high-level feature, so packing and padding are a bit overkill. It depends what the tuples themselves are backed by though. Actual, honest-to-SourceMod handles would require that they get closed without explicit calls to CloseHandle. I'm not sure I want to die on that hill right now. It's possible you meant some other kind of handle though.

dvander:
Thinking about this more: this PR was intended to paper over bad syntax, not bring a fully-featured implementation. So that makes me think:

 - Calling them "enum structs" (or something other than pure "structs") is better.
 - We should keep the behavior of sizeof(variable) (which this patch does) and sizeof(EnumStructType) (which is a TODO).
 - We should expose something like sizeof(EnumStructType.field) and something else to get the offsets - maybe just EnumStructType.field.

... Rather than going for gold and figuring out how to integrate true structs with containers like ArrayList.

dvander:
The latest patch adds support for @TotallyMehis's use case, using the syntax in the test case. I picked the resolution operator over the field operator, since it is really accessing meta-information and not a named static property.

assyrianic:
@dvander to be exact, my tuple implementation is a static, dynamically allocated array that uses a vector to store the fields and size of each member but this kind of tuple was implemented with C's limitations in design so it's likely very useless here unless you find a way to add type-safety to it.

Getting back on topic though, I think the `enum struct` is a good hold over in terms of having a way to cleanly organize data without having to resort to a weird combination of semantics and arrays.

Realistically, IIRC, if methodmaps can be backed with arrays in some way rather than with enums, then you've pretty much added practical classes.

VoiDeD:
Just a note per IRC to add tests for `sizeof(x.y)` and `sizeof(X::y)`.

dvander:
Thanks, that should work now.

compiler/parser.cpp
```diff
@@ -2671,6 +2671,11 @@ static int consume_line()
   return TRUE;
 }
 
+static cell char_array_cells(cell size)
```
Headline: if we're c++17 we can `constexpr` :)

dvander: Done!

assyrianic:
quick suggestion.
perhaps it's better to name it as `record` rather than `enum struct`?
```c
enum struct Foo {
    int x;
    int y;
    float z;
};
```
to
```c
record Foo {
    int x;
    int y;
    float z;
};
```

dvander:
This is syntactic sugar for a feature everyone already refers to as "enum structs". We're not really adding structs, or something analogous to structs, because enum structs compose differently and can be casted to/from arrays. I would rather call it what it is.

(A better name might be "array structs" or "named arrays", but I'm working with the keywords available.)

dvander:
I'm considering this PR feature-complete. I will probably have some follow-up commits to clean up parser support or fix bugs, but will otherwise leave it open for a few days for additional comments.

assyrianic:
Question. From the [SM wiki](https://wiki.alliedmods.net/SourcePawn_Transitional_Syntax#Caveats)
```7. Methodmaps can only be defined over scalars - that is, the "this" parameter can never be an array. This means they cannot be used for enum-structs.```

I'm going to assume that this limitation still persists with `enum struct` currently but is there really no way to allow `this` to be an array so that `enum struct`s can be applicable to a methodmap?

From one of the test code, I see that the methodmap is defined and then used by the enum struct when the usual way would be to define the enum struct and then the methodmap wraps around it.

dvander:
I'm tentatively planning to enable methods on enum structs, but probably not in this PR.

dvander:
One change I haven't decided on yet: whether to disallow enum struct types in natives. If two plugins use different versions of the same named enum struct, the offsets may be different. True structs would be able to deal with this, but right now, we can't, and it may lead to very obscure bugs if users don't version cross-plugin APIs properly.

It would be trivial to get around such a restriction by using "any", as ArrayList and friends already do.

PR #293, opened 2018-12-15
Disallow reference tokens on enum struct parameters.

dvander:
This essentially guarantees that future struct implementations will have
pass-by-reference semantics. The alternative would be requiring a
reference token for enum structs, and then leaving the door open for
pass-by-value semantics later. However, there do not seem to be many use
cases for pass-by-value that could not be solved in other ways. And this
implementation matches how arrays and methodmaps are already used.

### DIFF
--- a/compiler/messages.h
+++ b/compiler/messages.h
@@ -160,7 +160,7 @@ static const char *errmsg[] = {
 /*133*/  "cannot coerce unrelated object types %s and %s\n",
 /*134*/  "type mismatch (%s and %s)\n",
 /*135*/  "cannot use enum struct type \"%s\" in natives\n",
-/*136*/  "unused136\n",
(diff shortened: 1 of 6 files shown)


assyrianic:
Question, why have this as an error and not as a warning? I would imagine since the enum-struct is just an array, the `&` token would just be thrown out.

dvander:
Two reasons...

1. [Be liberal in what you accept](https://en.wikipedia.org/wiki/Robustness_principle) does not apply to programming language syntax.
2. When we introduce true structs, either (a) they will be pointer-like or (b) they won't. If (a), then suddenly we'll have lost the ability for `&` to indicate an outparam for a pointer. If (b), then it would be hugely confusing to have the behavior be different from existing structs, so we'd rather have *forced* the ampersand on enum structs to leave pass-by-value an option.

PR #296, opened 2018-12-18
Fix incorrect name output in error 135

### LINKED ISSUE #295: Error 135 uses wrong string in message
The following code:
```sourcepawn
native void Touhou_GetUserBullet(int client, bullet_t bullet);
```
results in the following error message:
```error 135: cannot use enum struct type "Touhou_GetUserBullet" in natives```

The native name is being used in the error message rather than the enum struct type name.

### PR DESCRIPTION
**Pull request by Headline**
Fixes #295

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -4952,9 +4952,10 @@ static int declargs(symbol *sym, int chkshadow, const int *thistag)
       if (decl.name[0] == PUBLIC_CHAR)
         error(56, decl.name); /* function arguments cannot be public */
 
-      if (gTypes.find(decl.type.semantic_tag())->isEnumStruct()) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: compiler/parser.cpp [hunk 1]
```diff
@@ -4952,9 +4952,10 @@ static int declargs(symbol *sym, int chkshadow, const int *thistag)
       if (decl.name[0] == PUBLIC_CHAR)
         error(56, decl.name); /* function arguments cannot be public */
 
-      if (gTypes.find(decl.type.semantic_tag())->isEnumStruct()) {
+	  Type *type = gTypes.find(decl.type.semantic_tag());
```
asherkin: Indentation looks off here.
```suggestion
 	Type *type = gTypes.find(decl.type.semantic_tag());
```

Issue #297, opened 2018-12-18
spcomp assertion on malformed if


Headline:
thanks to shavit for reporting this one.

Gave it a shot myself, but a solution wasn't obvious to me. Here's how to replicate

```
int main(){
    if(
    {
        int b = c();
    }
}
```
```
C:\Users\Micha\Desktop\alliedmodders\sourcemod\sourcepawn\build\compiler\spcomp>spcomp test.sp
SourcePawn Compiler 1.10
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2018 AlliedModders LLC

Assertion failed: lval1->sym!=NULL, file C:\Users\Micha\Desktop\alliedmodders\sourcemod\sourcepawn\compiler\expressions.cpp, line 1057
```

dvander:
Oh gross.

PR #299, opened 2018-12-20
Fix spcomp assertion on malformed if (#297)

dvander:


tests/compile-only/fail-broken-if-conditional.txt
```diff
@@ -0,0 +1 @@
+(6) : error 054: unmatched closing brace ("}")
```
Headline: this isn’t error `142`, is this intended?

dvander: To avoid error spam, the compiler only reports the first error on a line, and then it silences errors up until the next statement. So error 142 gets reported, but another error gets there first. (Arguably I should have re-used 29 since we're running out of error codes, but whatever.)

This is not ideal from a testing perspective but the parser is so incredibly brittle that not crashing is probably the best we can hope for.

PR #300, opened 2018-12-20
Fix infinite loop when parsing broken enum structs. (fixes #298)

### LINKED ISSUE #298: compiler cpu usage goes 100% with strange use of enum struct
`#include <sourcemod>

enum struct test{

  int test2;
  enum_struct_inside_enum_struct test3;

}

enum struct enum_struct_inside_enum_struct

public void OnPluginStart(){

  test testVar;
  testVar.test3();
}`

when i compile this with latest compiler, compiler doesn't show up anything and its process use 100% cpu. sorry for weird variable keyword

(tested on Windows, Mac)

dvander:


tests/testutil.py
```diff
@@ -31,15 +32,26 @@ def __exit__(self, exc_Type, exc_value, traceback):
     if self.cwd is not None:
       os.chdir(self.cwd)
 
-def exec_argv(argv):
+def exec_argv(argv, logger = None):
   if argv[0].endswith('.js'):
     argv = ['node'] + argv
 
   p = subprocess.Popen(argv, stdout = subprocess.PIPE, stderr = subprocess.PIPE)
```
Headline: Mentioned in IRC; but I'll throw this here. 

Popen has a timeout parameter that throws an Exception when that timeout is reached. Perhaps that's better?

dvander: Agree it would be better, but I think that's only available in a totally separate, unrelated language called Python 3.

PR #316, opened 2019-02-04
Fix crash when enabling profiler

peace-maker:
When a plugin is currently executing code while the profiler is being activated, the VM tried to leave a profiling scope it never entered.

`sm_rcon sm prof start vprof` would crash the server.

https://crash.limetech.org/lgolraffyt46

### DIFF
--- a/vm/environment.h
+++ b/vm/environment.h
@@ -209,15 +209,20 @@ class EnterProfileScope
  public:
   EnterProfileScope(const char* group, const char* name)
   {
-    if (Environment::get()->IsProfilingEnabled())
+    if (Environment::get()->IsProfilingEnabled()) {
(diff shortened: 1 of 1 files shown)


vm/environment.h
```diff
@@ -209,15 +209,20 @@ class EnterProfileScope
...
 
   ~EnterProfileScope()
   {
-    if (Environment::get()->IsProfilingEnabled())
+    if (scope_entered_ && Environment::get()->IsProfilingEnabled())
       Environment::get()->profiler()->LeaveScope();
   }
+
+ private:
+  boolean scope_entered_;
```
asherkin:
```suggestion
  bool scope_entered_;
```

PR #322, opened 2019-03-29
KE_CONSTEXPR define was removed

kgns:
KE_CONSTEXPR define was removed in this commit: https://github.com/alliedmodders/amtl/commit/7586906494fd1df46c468309798b5d3f49c95b39

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -2684,7 +2684,7 @@ static int consume_line()
   return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


peace-maker:
The amtl library is still pinned to an older commit in this repository, so that change doesn't affect sourcepawn code yet. It'll have to be made once a newer version of amtl is pulled in.

dvander:
Should be okay to take this now, anyway.

Issue #327, opened 2019-04-26
enums structs with char[] members are incorrectly sized


maximsmol:
**To reproduce:**
1. Define an `enum struct` with a character array member.
1. Print its `sizeof`/look at the assembly

**Example of faulty input:**
`test.sp`:
```cpp
enum struct Test {
  char a[10]; // 3 * cell_t = 12 * byte (rounded up) = 10 * byte
} // 3 * cell_t

public main() {
  Test test;
  printnum(sizeof(test)); // 10
  // ^ native from vm/shell.cpp
}
```

**Expected:**
`sizeof` is correct (3 for the example).
Allocated stack size is correct (12 for the example).

**Actual:**
`sizeof` is incorrect (10 for the example).
Allocated stack size is incorrect (40 for the example).

`test.asm`:
```
fill 28 ; <-- allocates 0x28 = 40 bytes = 10 * cell_t
; line 17
break ; dc4
const.pri a
push.pri
;$par
sysreq.n 2 1	; printnum
;$exp
stack 28
```

Fyren:
We do know about a bunch of string-related problems in `sizeof`/`cellsof` (not only with enum structs), but they all boil down to how strings were hacked into SP in the first place.  Original Pawn only (I think) considered sizes in cells, but most of the modified spots for strings consider bytes.  When the two intersect, like sometimes with char arrays or with them in enum structs, there's confusion and sizes in bytes get treated as cells.  This is probably not impossible to fix, but may be somewhat deep-reaching.  If there's a correctness problems, we'd definitely make an effort.

maximsmol:
Sounds reasonable. I am not really invested in getting this fixed for spcomp1, but it came up in my process of studying spcomp1 behavior to properly implement `enum struct`s in spcomp2 (exp/compiler). I asked dvander whether this is a bug or if I am misunderstanding something, and he confirmed that this is not intentional and asked me to file an issue for him, so here we are. :)

dvander:
I'm interested in fixing this for transitional enum structs, while we have the chance. Old ones not so much. Might even break compatibility somehow, like if the struct is communicated across plugins and the size is computed differently.

PR #328, opened 2019-04-27
Enum struct sizeof with string (#327)

Fyren:
This only changes new-style enum structs, but the change for the old style ought to be basically the same.

I'm unsure if there's some new test that would be relevant.  I updated the existing ones that assumed the old behavior.  This _is_ a behavior change, but I suspect it won't impact anyone.

I'll change old-style enums if we want that as well.


dvander:
lgtm

PR #331, opened 2019-05-07
Fix incorrect reported file in error message when missing a required semicolon.

Fyren:
This would only happen with `#pragma semicolon 1` if a semicolon was missing after some other file was included.

Without the change, the error reported for the test case would be in the empty include (though with the correct line number).


peace-maker:
Are you sure the test is correct? You're including a nonexistent file. Maybe I'm not understanding the bug here.

Fyren:
I'm including an almost empty file.  Try the test against master (i.e. without this PR) and the error reported will be in the empty file.

Almost empty because apparently if you try to include an actual 0-byte file, spcomp will error.

asherkin:
> I'm including an almost empty file. Try the test against master (i.e. without this PR) and the error reported will be in the empty file.

I think it got renamed at one point, the filename here does indeed seem wrong.

Fyren:
I did rename it, but the test still passed because the check was lax.  Fixed, thanks.

Issue #333, opened 2019-05-07
Segfault with invalid methodmap declaration


Sikarii:
Following code seems to segfault both the 1.9 and 1.10(?) compilers:

### Test case
```
methodmap break TestingA < StringMap
{
    public TestingA()
    {
        return view_as<TestingA>(new StringMap());
    }

    public void Test()
    {
        this.SetValue("test", 0);
    }
}

methodmap TestingB < TestingA
{
    public TestingB()
    {
        return view_as<TestingB>(new TestingA());
    }
}

TestingB g_Test = null;

public void OnPluginStart()
{
    g_Test = ReturnB();
}

TestingB ReturnB()
{
    return new TestingB();
}

public void OnConfigsExecuted()
{
    g_Test.Test();
}
```

---
### Output
```
SourcePawn Compiler 1.9.0.6280
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2017 AlliedModders LLC

Segmentation fault

SourcePawn Compiler 1.10.0.6405
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2018 AlliedModders LLC



```

### Expected output
```
plugin.sp(1) : error 093: expected tag name identifier
```

Notice how the 1.10 compiler does not report segmentation fault, but I highly suspect it does so.
Also worth noting that this does not segfault when on https://spider.limetech.io/ (older version?)

Fyren:
It's definitely smashing memory.  Whether that results in a segfault or terminating in some other way is probably just up to chance.  Spider will act differently in cases like this because it's JS running in your browser.  

This specific problem seems to be in the methodmap parsing at https://github.com/alliedmodders/sourcepawn/blob/5e1091f923880662466634a0794910dde104c61e/compiler/parser.cpp#L3652.  Not doing so appears to end up causing problems down the line.  Going to wait on pushing a change until checking into if the same problem exists in other places.

PR #334, opened 2019-05-11
Ensure that methodmaps always have a name.

dvander:
When we encounter an unknown method call, we will try to generate a
temporary symbol with a made-up name. If the methodmap has no name,
this will result in a symbol starting with ".", for example,
".SetValue". When finding the display name for this function, the
compiler assumes that a name starting with a non-identifier symbol
is special, and must therefore be an operator. Since operator names have
a different maximum length than normal symbol names, this causes a stack
overflow.

In addition this patch fixes the function that can erroneously overflow.

Bug: fixes issue #333
Test: compile-only/fail-methodmap-keyword-name

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -31,6 +31,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
+#include <amtl/am-string.h>
 #include <amtl/am-unused.h>
 #include <amtl/am-platform.h>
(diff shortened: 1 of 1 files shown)


peace-maker:
Did you intend to add a test?

KyleSanderson:
Fixed in https://github.com/alliedmodders/sourcepawn/pull/335 . Thanks P-M.

PR #337, opened 2019-05-26
Remove pointless comments.

dvander:
This removes the following comments:
 - /* do */
 - /* for */
 - /* if */
 - /* switch */
 - /* while */


Fyren:
```
#define IF if(
#define THEN ){
#define ELSE } else {
#define ELIF } else if (
#define FI ;}
```

PR #340, opened 2019-06-08
Restyle spcomp.

dvander:
This restyles include/*.h and compiler/*.{h,cpp} in AMTL style.


compiler/amxdbg.h
```diff
@@ -28,146 +28,150 @@
...
+    int16_t states PACKED;     /* number of entries in the "state" table */
 } PACKED AMX_DBG_HDR;
-#define AMX_DBG_MAGIC   0xf1ef
+#define AMX_DBG_MAGIC 0xf1ef
 
 typedef struct tagAMX_DBG_FILE {
-  ucell   address       PACKED; /* address in the code segment where generated code (for this file) starts */
-  char    name[1];           /* ASCII string, zero-terminated */
+    ucell address
+        PACKED;   /* address in the code segment where generated code (for this file) starts */
```
Fyren: Ergh.

compiler/assembler.cpp
```diff
@@ -64,1326 +64,1322 @@ class CellWriter;
...
-  {135, "sysreq.n",   sIN_CSEG, parm2 },  /* version 9 (replaces SYSREQ.d from earlier version) */
-  {161, "tracker.pop.setheap", sIN_CSEG, parm0 },
-  {160, "tracker.push.c", sIN_CSEG, parm1 },
-  { 35, "xchg",       sIN_CSEG, parm0 },
-  { 83, "xor",        sIN_CSEG, parm0 },
-  { 91, "zero",       sIN_CSEG, parm1 },
-  { 90, "zero.alt",   sIN_CSEG, parm0 },
-  { 89, "zero.pri",   sIN_CSEG, parm0 },
-  { 92, "zero.s",     sIN_CSEG, parm1 },
+    /* node for "invalid instruction" */
```
Fyren: This is probably better before.

compiler/amxdbg.h
dvander: Thanks, will fix this.

compiler/assembler.cpp
dvander: Agree, I'll disable clang-format around this list.

compiler/expressions.h
```diff
@@ -33,50 +33,48 @@ struct svalue;
...
+    int hier9(value *lval);
+    int hier8(value *lval);
+    int hier7(value *lval);
+    int hier6(value *lval);
+    int hier5(value *lval);
+    int hier4(value *lval);
+    int hier3(value *lval);
+    int hier2(value *lval);
+    int hier1(value *lval1);
+    int plnge(int *opstr, int opoff, HierFn hier, value *lval, const char *forcetag,
```
Fyren: I'm not sure what the deal is, but it seems like where the asterisk gets nestled by the formatter varies.  Here, it's with the variable name and in other places its with the type.

Example in `compiler/assember.h`: `void assemble(const char* outname, memfile_t* fin);`

dvander: I think I have to forcefully disable this in the clang-format config.

dvander:
no more shrieks of horror, let's do this

PR #345, opened 2019-07-22
Simplify inline exit frames.

dvander:
When invoking C++ from JIT code, we need to construct an "exit frame"
so that FrameIterator can recover a stack trace. The normal way of
creating an exit frame is implicitly, via a call. The call pushes a
return address, and the new prologue configures the frame pointer and
stores it in the Environment.

Notably, this means there is a small inline helper function every time
we transition to C++. For performance-sensitive code, like native calls,
this is not ideal. Instead we create an "inline" exit frame. Instead of
calling, we derive the return address at compile-time, push it, and thus
avoid the call instructions.

This patch makes additional improves to the inline-exit path by also
eliding "leave" instructions. It also preserves ebp, which should
improve stack walking by Breakpad.

### DIFF
--- a/vm/jit.h
+++ b/vm/jit.h
@@ -130,6 +130,7 @@ class CompilerBase : public PcodeVisitor
   Label throw_error_code_[SP_MAX_ERROR_CODES];
   Label report_error_;
   Label return_reported_error_;
+  Label unbound_native_error_;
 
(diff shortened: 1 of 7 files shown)


dvander:
re: x64 - no JIT yet, so I just removed some unused functions.

Issue #349, opened 2019-08-18
Warning is generated when calling functions outside enum structs from enum struct functions


naydef:
# Help us help you
 
     * [x]  I have checked that my issue [doesn't exist yet](https://github.com/alliedmodders/sourcemod/issues).
 
     * [x]  I have tried my absolute best to reduce the problem-space and have provided the absolute smallest test-case possible.
 
     * [x]  I can always reproduce the issue with the provided description below.
 
 
 # Environment
 
     * Game/AppID (with version if applicable): -
 
     * Current SourceMod version: 1.10.0.6434
 
     * Current Metamod: Source snapshot: -
 
 
     * [x]  I have updated SourceMod to the [latest version](https://www.sourcemod.net/downloads.php) and it still happens.
 
     * [x]  I have updated SourceMM to the [latest snapshot](https://sourcemm.net/downloads.php?branch=dev) and it still happens.
 
 
 # Description
 
 Calling a function outside enum struct from enum struct function generates a warning.
 Also the error line doesn't match the corresponding file.
 ```
 //SourceMod Batch Compiler
// by the SourceMod Dev Team


//// id3v2pawn_test.sp
//
// C:\Users\Geo\Downloads\sourcemod-1.8.0-git5985-windows\addons\sourcemod\scrip
ting\id3v2pawn_test.sp(24) : warning 203: symbol is never used: "fun"
// Code size:             3068 bytes
// Data size:             2280 bytes
// Stack/heap size:      16384 bytes
// Total requirements:   21732 bytes
//
// 1 Warning.
//
// Compilation Time: 0.11 sec
// ----------------------------------------

Press enter to exit ...





```
 # Problematic Code (or Steps to Reproduce)
 
 id3v2pawn_test.sp
 ```
#include <sourcemod>
#include <splendid>

public void OnPluginStart()
{
	My_Struct bgm_info;
	bgm_info.size=0;
}
 ```
 
  splendid.inc
 ```
#if defined _SS_
 #endinput
#endif
#define _SS_












enum struct My_Struct
{
	int size;
	void File()
	{

		this.size=fun(this.size);
	}
	
}

int fun(int value)
{
    return value;
}

 ```

P.S. I'm sorry for the bad formatting, I've simply copied the formatting and the text from another issue.

peace-maker:
[...]

PR #352, opened 2019-08-23
Fix warning when non-stock functions are used by stock functions.

dvander:
Bug: issue #349
Test: warnings/unused-from-member-function.sp


dvander:
I think you're going that route, you might as well remove stock entirely. Its value is maximized when used sparingly, and that would encourage it to appear everywhere.

dvander:
Note: this makes it behave more like "static inline" in C++.

Issue #301, opened 2018-12-23
Can't get Angles with GetClientEyeAngles in enum struct.


Bara:
Here's the test case with compile error and outputs for the working methods:
https://pastebin.com/aXh5Nsiu

jason-e:
Chiming in to say: this issue also happens with old-style enum structs as well, it is not unique to the new syntax. Would be great to have it fixed.

dvander:
Thanks for the report. This should indeed be fixed, but it will be difficult (unless I can find a surgical hack). The root of the issue is that the compiler does not really have types internally, and the only way it can represent an array type is through the same data structure that represents variables.

This obviously poses a problem for multi-dimensional arrays, since you need a separate type to represent inner dimensions. The compiler works around this by creating a hidden variable for each extra dimension. For example, "int x[MAXPLAYERS][3]" will actually have two definitions of "x", one for the outer dimension, and a hidden one for the inner.

How does this relate to enum structs? Well, enum structs don't allow multi-dimensional arrays, so the compiler doesn't bother creating symbols for embedded arrays. Instead it uses some really awful hacks to pretend there's a symbol there, just long enough to do some immediate type checking. Then it's gone, and the expression devolves to an anonymous array which does not have size information.

By the time the compiler is checking your function call, the size information is long gone.

I can think of two fixes for this. The difficult fix is also the proper one: remove the ugly hack, and create hidden symbols for array types within enum structs. The hack is ancient, I would only feel comfortable removing it for modern enum structs. The quicker solution would be to propagate the size through the `value` mechanism in the expression parser, so the argument checker can read it. But that is an extra smearing of hacks on a towering layer cake of hacks, and I'm hoping not to do that.

PR #353, opened 2019-08-24
Create actual symbols for enum struct fields.

dvander:
hier1() had an ancient hack that created a dummy symbol for enum struct
fields. This dummy symbol was stack allocated, so we couldn't propagate
it through the expression parser. This attaches the symbols to the
variable so they can escape, which gives the compiler much richer
information about the field. Note that these one-off symbols could be
re-used, but we do not re-use them in this patch.

Bug: issue #301
Test: tests/enum-structs


dvander:
The original code here was super hacky, and I was never confident in it. We have a lot of tests around it because it was so buggy during development. The new code passes those tests, and the test case from the reported issue - but I'm still not 100% confident.

`symbol` has two duties - it contains the normal properties you'd expect of a variable (type, address, name), but it also pulls in parts of the type that are not expressible by tags. This means that it is *necessary* to have a symbol in some cases, because the tag field of `lval` isn't enough to do type checking. This is why, for example, string literals sometimes don't work as arrays.

In this patch, we're creating a symbol out of nothing, and faking just enough information so that type checking will work. We're relying on the properties of the lval (it'll always be a REFARRAY or ARRAYCELL) to guarantee we'll never call something like symbol::addr(). It's a bit hokey, but let's see where it goes.

PR #364, opened 2019-09-09
Disqualifiy const from enum struct fields

### LINKED ISSUE #363: Enum Struct: const members
SP allows enum struct members to have const keyword, although it has no effect.

Unless there's some way/plan to initialize const members, maybe prevent the keyword from being used.
```
#include <sourcemod>

enum struct Test {
	const int x;
}

public void OnPluginStart() {
	RegAdminCmd("sm_testConst", cmdTestConst, ADMFLAG_ROOT);
}

public Action cmdTestConst(int client, int args) {
	Test t;
	t.x = 3;
	ReplyToCommand(client, "%i", t.x);
	return Plugin_Handled;
}
```

Headline:
Fixes #363 

Prevents enum struct fields from being `const`. I'm not sure we want to support const enum struct members, so if that is something we want I can refactor this, but I went straight for removal & error.

Added test

### DIFF
--- a/compiler/messages.h
+++ b/compiler/messages.h
@@ -118,7 +118,7 @@ static const char* errmsg[] = {
     /*091*/ "ambiguous constant; tag override is required (symbol \"%s\")\n",
     /*092*/ "number of arguments does not match definition\n",
     /*093*/ "expected tag name identifier\n",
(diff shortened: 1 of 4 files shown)


dvander:
Thanks for the test!

PR #368, opened 2019-09-11
Remove uses of PAWN_CELL_SIZE.

dvander:
Bug: n/a
Test: manual test

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -659,43 +659,9 @@ ftoi(cell* val, const unsigned char* curptr)
         error(70); /* rational number support was not enabled */
         *val = 0;
     } else if (rational_digits == 0) {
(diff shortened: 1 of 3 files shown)


compiler/lexer.cpp
```diff
@@ -3323,17 +3289,7 @@ itoh(ucell val)
...
-#if PAWN_CELL_SIZE == 16
-    max = 4;
-#elif PAWN_CELL_SIZE == 32
-    max = 8;
-#elif PAWN_CELL_SIZE == 64
-    max = 16;
-#else
-#    error Unsupported cell size
-#endif
+    int max = 8;
```
Fyren: Oh god, dvander.  How can you waste all that memory in the `nibble` array like that?

Issue #370, opened 2019-09-11
warning 219: local variable "i" shadows a variable  at a preceding level


naydef:
**Min example (rtest.sp):**
```
#pragma semicolon 1
#pragma newdecls required

#include <sourcemod>


public Action Timer_AgainCanSee(int s)
{
	DoOverlay(i);
	return Plugin_Continue;
}

void DoOverlay(int client)
{
	IsClientInGame(client);
}
```

Compile log (SM 1.10):
```
//SourceMod Batch Compiler
// by the SourceMod Dev Team


//// rtest.sp
//
// C:\Users\Geo\Downloads\sourcemod-1.8.0-git5985-windows\addons\sourcemod\scrip
ting\include\menus.inc(368) : warning 219: local variable "i" shadows a variable
 at a preceding level
// C:\Users\Geo\Downloads\sourcemod-1.8.0-git5985-windows\addons\sourcemod\scrip
ting\rtest.sp(9) : error 017: undefined symbol "i"
//
// 1 Error.
//
// Compilation Time: 0.19 sec
// ----------------------------------------

Press enter to exit ...







```

The first message shouldn't be printed. It's interesting to note that the compiler complains about a local variable in a member function in a methodmap. Replacing the _i_ character doesn't generate the warning message and everything works correctly. This issue caused me headache(took me long to find the cause of many, many errors and warnings printing for some strange reason)

dvander:
It's on my todo list to fix this, but... it's very difficult to do properly. The problem is that undefined variables will implicitly create a global function, which if you use a common name like "i" will almost certainly create an erroneous aliasing message.

I could potentially hack around this particular case by omitting the warning if the symbol is a function but not implemented (i.e. only has the implicit prototype).

Headline:
> The problem is that undefined variables will implicitly create a global function, which if you use a common name like "i" will almost certainly create an erroneous aliasing message.

Is this legacy? Can't we rip this out

dvander:
I've been looking for ways to remove it but it's not been easy. It's critical to the two-pass compilation process. Functions are the only things that can be out of order, so the compiler assumes anything undefined must be a function defined later in the file, and then pre-populates the symbol. No matter how it gets changed, it will involve a fair bit of surgery.

dvander:
For those curious, the prepopulation happens in a call to `fetchfunc`, which implictly adds a new symbol.

PR #371, opened 2019-09-12
Relax name shadowing checks for implicitly prototyped functions.

dvander:
Bug: issue #370
Test: manual test


dvander:
Can't figure out how to write a good test this one, the harness doesn't have any way to expect *no* warning, and any test case for this will fail to compile.

PR #372, opened 2019-09-14
Remove the iLABEL identifier type.

dvander:
This has never been used, and we don't support goto anyway.

Bug: N/A
Test: builds

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -349,7 +349,7 @@ pc_compile(int argc, char* argv[]) {
     do {
         /* reset "defined" flag of all functions and global variables */
         reduce_referrers(&glbtab);
-        delete_symbols(&glbtab, 0, TRUE, FALSE);
+        delete_symbols(&glbtab, 0, FALSE);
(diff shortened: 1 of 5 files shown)


compiler/parser.cpp
```diff
@@ -5670,10 +5660,10 @@ compound(int stmt_sameline) {
 
     testsymbols(&loctab, nestlevel, FALSE, TRUE); /* look for unused block locals */
     declared = save_decl;
-    delete_symbols(&loctab, nestlevel, FALSE, TRUE); /* erase local symbols, but
-                                                      * retain block local labels
-                                                      * (within the function) */
-    nestlevel -= 1;                                  /* decrease compound statement level */
+    delete_symbols(&loctab, nestlevel, TRUE); /* erase local symbols, but
+                                               * retain block local labels
+                                               * (within the function) */
```
Fyren: This comment is now useless.  It is extremely important that we remove it.

dvander: Done!

PR #373, opened 2019-09-16
Remove legacy enum structs.

dvander:
This patch effectively cuts enum structs off at the knees. The root constant
created by an enum is now defined as 0, and it's illegal to create
arrays using this constant. Parsing support has been removed from the
index operator, and the index operator (and old-array decl syntax) now
require types implicitly convertible to integers.

In addition, support for tagging items in enums has been removed, as
there is no longer a need for this behavior.

The underlying mechanism for this mess is still supported, however,
because it is used for transitional enum structs. This patch simply
removes the worst hacks in the parser and hier1().

Plugins using the old array-as-struct syntax will no longer compile, and
should use the new enum struct syntax instead.

Bug: N/A
Test: manual test


psychonic:
⚰️

compiler/expressions.cpp
```diff
@@ -1896,6 +1896,18 @@ SC3ExpressionParser::parse_view_as(value* lval)
     return lvalue;
 }
 
+bool
+is_valid_index_tag(int tag)
+{
+    Type* idx_type = gTypes.find(tag);
+    if (idx_type->isStruct() || idx_type->isFunction() || idx_type->isEnumStruct() ||
+        idx_type->isMethodmap() || tag == pc_tag_void || tag == sc_rationaltag)
```
Fyren: `pc_tag_string`?  I dunno if there's a better way than listing these all out.

compiler/expressions.cpp
```diff
@@ -2062,39 +2066,11 @@ SC3ExpressionParser::hier1(value* lval1)
...
-            } else {
-                assert(sym != NULL);
-                if (cursym != &dummysymbol)
-                    lval1->tag = sym->tag;
-                lval1->constval = 0;
-            }
+
+            assert(sym != NULL);
+            if (cursym != &dummysymbol)
+                lval1->tag = sym->tag;
```
Fyren: This is dead now and `dummysymbol` is never used.

Fyren:
Tests that trigger the new errors would be nice.

compiler/expressions.cpp
dvander: I couldn't decide if pc_tag_string should be included, indexing by char sounds like a legit thing.

compiler/expressions.cpp
dvander: Thanks, it gives me great joy to remove this declaration.

dvander:
Added tests.

compiler/messages.h
```diff
@@ -177,7 +177,7 @@ static const char* errmsg[] = {
     /*150*/ "setter must take exactly one extra argument with type %s\n",
     /*151*/ "unmatched opening brace ('{') (line %d)\n",
     /*152*/ "no setter found for property %s\n",
-    /*153*/ "unused153\n",
+    /*153*/ "Arrays as enums have been deprecated. See https://wiki.alliedmods.net/SourcePawn_Transitional_Syntax#Enum_Structs\n",
```
Headline: I’m not sure if we should use the wording “deprecated”, maybe “removed?”

dvander:
Improved error message.

PR #377, opened 2019-09-24
Forbid passing character slices to references.

dvander:
This can't be properly supported because there is no 8-bit char type. As
is, it performs an improper memory access.

Bug: N/A
Test: compile-only/fail-array-char-to-int-ref


dvander:
Note: upstream contains the same fix.

PR #395, opened 2019-10-23
Switch to the AMTL argument parser.

dvander:
This removes a bunch of hand-rolled argument parsing code. Two options
have been removed: -^ and -\ which allowed switching the control
character, generally a poor idea for having shareable code.

### DIFF
--- a/compiler/parser.cpp
+++ b/compiler/parser.cpp
@@ -23,9 +23,6 @@
  *
  *  Version: $Id$
  */
-#include <amtl/am-platform.h>
-#include <amtl/am-string.h>
-#include <amtl/am-unused.h>
 #include <assert.h>
 #include <ctype.h>
 #include <errno.h>
@@ -34,6 +31,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+#include <amtl/am-platform.h>
+#include <amtl/am-string.h>
+#include <amtl/am-unused.h>
+#include <amtl/experimental/am-argparser.h>
+
 #include "types.h"
 
 #if defined __WIN32__ || defined _WIN32 || defined __MSDOS__
@@ -95,6 +98,8 @@ static unsigned total_drives; /* dummy variable */
 #define VERSION_STR "3.2.3636"
 #define VERSION_INT 0x0302
 
+using namespace ke;
+
 int pc_anytag = 0;
 int pc_functag = 0;
 int pc_tag_string = 0;
@@ -115,7 +120,6 @@ static char* get_extension(char* filename);
 static void setopt(int argc, char** argv, char* oname, char* ename, char* pname);
 static void setconfig(char* root);
 static void setcaption(void);
-static void about(void);
 static void setconstants(void);
 static void parse(void);
 static void dumplits(void);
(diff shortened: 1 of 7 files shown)


compiler/parser.cpp
```diff
@@ -717,189 +721,182 @@ set_extension(char* filename, const char* extension, int force) {
...
-                    if (*(ptr + 1) != '\0')
-                        about();
-                    sc_asmfile = TRUE; /* skip last pass of making binary file */
-                    if (verbosity > 1)
-                        verbosity = 1;
-                    break;
-                case 'D': /* set active directory */
-                    ptr = option_value(ptr, argv, argc, &arg);
+
+    auto usage = Sprintf("[options] <filename> [filename...]", argv[0]);
```
peace-maker: That `argv[0]` argument seems unused.

compiler/parser.cpp
```diff
@@ -717,189 +721,182 @@ set_extension(char* filename, const char* extension, int force) {
...
+    if (opt_error_file.hasValue())
+        strlcpy(ename, opt_error_file.value().chars(), _MAX_PATH);
+
 #if defined __WIN32__ || defined _WIN32 || defined _Windows
-                case 'H':
-                    hwndFinish = (HWND)atoi(option_value(ptr, argv, argc, &arg));
-                    if (!IsWindow(hwndFinish))
-                        hwndFinish = (HWND)0;
-                    break;
+    hwndFinish = (HWND)atoi(opt_hwnd.value().chars());
```
peace-maker: This crashes on windows when not passing the `-H` option due to not having a default value.

compiler/parser.cpp
```diff
@@ -1069,60 +1067,6 @@ setcaption(void) {
     pc_printf("Copyright (c) 2004-2018 AlliedModders LLC\n\n");
 }
 
-static void
-about(void) {
-    if (strlen(errfname) == 0) {
-        setcaption();
-        pc_printf("Usage:   spcomp <filename> [filename...] [options]\n\n");
```
peace-maker: The old documentation has filenames first then options. The new usage output shows options first. I guess the old way around still works too?

compiler/parser.cpp
```diff
@@ -1069,60 +1067,6 @@ setcaption(void) {
...
-        pc_printf("         -v<num>  verbosity level; 0=quiet, 1=normal, 2=verbose (default=%d)\n",
-                  verbosity);
-        pc_printf("         -w<num>  disable a specific warning by its number\n");
-        pc_printf("         -z<num>  compression level, default=9 (0=none, 1=worst, 9=best)\n");
-        pc_printf("         -E       treat warnings as errors\n");
-        pc_printf("         -\\       use '\\' for escape characters\n");
-        pc_printf("         -^       use '^' for escape characters\n");
-        pc_printf("         -;<+/->  require a semicolon to end each statement (default=%c)\n",
-                  sc_needsemicolon ? '+' : '-');
-        pc_printf("         sym=val  define constant \"sym\" with value \"val\"\n");
```
peace-maker: All this extra documentation is gone.

compiler/parser.cpp
dvander: Yes.

compiler/parser.cpp
dvander: I can add notes back for sym=val line.

compiler/parser.cpp
dvander: Thanks, good catch, will fix this.

dvander:
Is there any value to the "response file" thing? If not, I'd rather just remove it.

dvander:
Response files are removed.

Issue #402, opened 2019-10-30
SM 1.11 `void` function should return value if we compare them (without calling).


TheByKotik:
If you try do something like it you receive: `error 209: function "DoNothing" should return a value`.
I'm thinking problem in line 10 (`if ( s_fnCallback != DoNothing )`).
```
#pragma newdecls required
#pragma semicolon 1

typedef Type = function void (const bool bSuccessful);
stock static Type s_fnCallback;

public void OnPluginStart ()
{
	bool bSuccessful;
	if ( s_fnCallback != DoNothing )
	{
		Call_StartFunction( null, s_fnCallback );
		Call_PushCell( bSuccessful );
		Call_Finish();
		s_fnCallback = DoNothing;
	}
}

stock static void DoNothing (const bool bSuccessful) {}

dvander:
Thanks, I can reproduce this and have a solution in mind. Hopefully will have a fix tonight.

PR #404, opened 2019-11-01
Fix marking function return values.

dvander:
The old parser could not determine whether a function call result was
used or not. To implement the warning, it would aggressively check
whether the inputs to an expression involved a function symbol, and mark
it as needed a return value. This mostly worked because check was copied
into the most common cases in the parser.

This broke in the new parser because functions identifiers are pipelined
rather than immediately evaluated. They always return a function symbol,
and the caller is responsible for evaluating it. This is different from
the old parser, where a function symbol immediately created a call
expression or a "function value" expression. As a result, the hack for
aggressively poking at potential function calls no longer works.

This can be resolved by using the AST. Each node must implement a
ProcessUses() call. For each child expression whose value is consumed,
it must call MarkUsed(). For CallExpr, this will mark the function as
needing a return value.

Bug: issue #402
Test: compile-only/ok-void-function-as-value.sp

### DIFF
--- a/compiler/semantics.cpp
+++ b/compiler/semantics.cpp
@@ -124,6 +124,11 @@ RvalueExpr::RvalueExpr(Expr* expr)
     }
 }
 
+void
+RvalueExpr::ProcessUses()
+{
+    expr_->MarkAndProcessUses();
+}
 
 bool
 IsDefinedExpr::Analyze()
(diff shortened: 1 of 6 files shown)


compiler/semantics.cpp
```diff
@@ -86,6 +86,26 @@ static inline OpFunc TokenToOpFunc(int token) {
     }
 }
 
+static void
+CheckFunction(const value* lval)
```
Headline: Not to be stingy, but for others reading this code `CheckFunction` leaves me asking "for what?"

dvander: If I knew the answer I'd tell you!

dvander: Okay, I see. It's serving double-duty, it's supposed to also check whether an expression is using a value but not returning one. In that case, I think we can eliminate it.

dvander:
Thanks for pushing on CheckFunction!

Issue #413, opened 2019-11-13
Assertion `!(sym->usage & uSKIPPED)' failed.


BotoX:
test.sp with only:
```
#include <multicolors>
```

hits the assert:
```alliedmodders/sourcemod/sourcepawn/compiler/assembler.cpp:424: void do_call(CellWriter *, AsmReader *, cell): Assertion `!(sym->usage & uSKIPPED)' failed.```

Remove the `#include <multicolors>` (empty file now) and it compiles fine.

Testcase: https://cloud.botox.bz/s/EKzHMNJzoJoHtR3
`./compile.sh`

dvander:
This is a bug, but blessedly not a serious one. The job of this assertion is to detect when we emit calls to functions that won't be assembled. The function making the call in this case is unreachable, so there's no chance of a code generation error. However the fact that we're generating code for unreachable functions is also a bug.

The compiler has a bunch of hacks to resolve cases where stocks call other stocks, but basically, these hacks do not work when stock1 calls stock2, which calls stock3, etc... and each call happens before the stock is declared.  The long-term fix for this is to replace parser.cpp. Trying to fiddle with it otherwise is very risky.

I think to make this warning go away, I will add a code generator hack to skip over lines in the assembly buffer belonging to an unreachable function.

dvander:
Oh, and - thanks for the great test case, I was able to reproduce this.

dvander:
I found a simpler fix - and it looks like this does not need a backport, so I'll close this when SourceMod takes the next SP update.

PR #417, opened 2019-11-24
Do not mark liveness information from skipped functions.

dvander:
Bug: issue #413
Test: manual test


Headline:
let’s pin this to sourcemod master asap

PR #418, opened 2019-11-25
Fix scalars being assignable to arrays.

dvander:
I truly do not understand the condition that was supposed to prevent this. Possibly, the original author meant the last "&&" to be an "||". But even taking that into account, it looks bogus. I can't think of anything illegal it actually prevents, except maybe constant literals that are <= 0.

### DIFF
--- a/compiler/semantics.cpp
+++ b/compiler/semantics.cpp
@@ -440,18 +440,15 @@ BinaryExpr::ValidateAssignmentRHS()
                        !oper_ &&
                        (right_val.ident == iARRAY || right_val.ident == iREFARRAY));
     if (leftarray) {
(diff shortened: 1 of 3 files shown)


Fyren:
The `constval` check is because it is actually used for arrays, at least sometimes:

https://github.com/alliedmodders/sourcepawn/blob/master/compiler/expressions.cpp#L3269-L3278
https://github.com/alliedmodders/sourcepawn/blob/master/compiler/expressions.cpp#L1546-L1554

But it doesn't look like `right_val` is tested to see if it's an array there.

dvander:
Yeah, indeed... it's hard to see how it's not bogus.

PR #421, opened 2019-11-25
exp::parser - generic json renderer outstream

rumblefrog:
## Summary of change

JsonRenderer constructor param was changed to a more generic out stream from std

## Change motivation

To allow the ability to store serialized JSON into a buffer and used else other (in my case, ffi), instead of stdout.

Docparse output has been linted with jq

### DIFF
--- a/exp/compiler/parser/json-tools.cpp
+++ b/exp/compiler/parser/json-tools.cpp
@@ -21,8 +21,8 @@
 using namespace ke;
 using namespace sp;
 
-JsonRenderer::JsonRenderer(FILE* fp)
- : fp_(fp),
+JsonRenderer::JsonRenderer(std::ostream& output)
+ : out_(output),
    indent_(0)
 {
 }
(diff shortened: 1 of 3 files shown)


exp/compiler/parser/json-tools.cpp
```diff
@@ -16,13 +16,15 @@
 // You should have received a copy of the GNU General Public License along with
 // SourcePawn. If not, see http://www.gnu.org/licenses/.
 #include "json-tools.h"
+#include <iostream>
+#include <sstream>
 #include <stdio.h>
```
Headline: is this still necessary? ditto on the header file

exp/compiler/parser/json-tools.cpp
```diff
@@ -21,8 +21,8 @@
 using namespace ke;
 using namespace sp;
 
-JsonRenderer::JsonRenderer(FILE* fp)
- : fp_(fp),
+JsonRenderer::JsonRenderer(std::ostream &output)
```
dvander: nit: `std::ostream& output` (& cuddles type, not name)

exp/compiler/parser/json-tools.h
```diff
@@ -45,14 +47,15 @@ class JsonObject;
 class JsonRenderer
 {
  public:
-  JsonRenderer(FILE* fp);
+  JsonRenderer(std::ostream &output);
```
dvander: ditto here

exp/compiler/parser/json-tools.h
```diff
@@ -45,14 +47,15 @@ class JsonObject;
...
 
   void Render(JsonValue* value);
 
-  FILE* fp() const {
-    return fp_;
+  std::ostream &stream() {
```
dvander: ditto here

exp/compiler/parser/json-tools.h
```diff
@@ -66,7 +69,7 @@ class JsonRenderer
   void prefix();
 
  private:
-  FILE* fp_;
+  std::ostream &out_;
```
dvander: ditto here

rumblefrog:
Should be addressed now.

dvander:
Thanks!

Issue #424, opened 2019-11-27
float/int addition issue


bakugo:
Consider the following code:

`PrintToChatAll("%f", (-1.0 - 1 + 1));`

When compiled with sourcemod 1.10, it prints `-1.000000`, the expected result
When compiled with sourcemod 1.11, it prints `-1073741824.000000` instead

dvander:
Thanks, I can reproduce this. I don't know the root cause yet but I'll look into it soon.

Fyren:
https://github.com/alliedmodders/sourcepawn/blob/0af421e3a4eb19c49d96b57061a6a4d86b9bb9a6/compiler/code-generator.cpp#L253

I believe this check isn't getting the right thing done.  We care about `operator+(float, int)`, which isn't commutative, but we're only looking at `ob_add`.

dvander:
Thanks, that is indeed the issue... fix incoming, will have a more in-depth writeup in the commit message.

dvander:
Thanks, the fix for this is now in SourceMod.

PR #428, opened 2019-12-02
Fix constants being handled incorrectly with commutative operator overloads.

dvander:
The old expression parser would actually swap value structs here, so
find_userop would implicitly swap parameters. This detail did not carry
into the new parser, because the semantic and codegen steps are now
separate. The semantic analysis step determines the overload signature,
and by the time we reach code generation, it is too late to pretend the
semantic values are swapped (this would require choosing a new
overload).

A quick fix is to force swapparams=true on the UserOperation when we make
this optimization. Longer term, this logic will be replaced. The
experimental compiler has a bunch of improvements to how PRI/ALT are used
that we want to backport to the production compiler.

Bug: issue #424
Test: tests/basic/constval-user-op


dvander:
Yeah, this is a good example of why code coverage is a wildly incomplete metric. Both the constval and the user operation code was covered, but not both together.

PR #429, opened 2019-12-02
Error if an enum struct method is directly accessed and not an actual call (fixes #406)

### LINKED ISSUE #406: Enum Struct function - No warning on missing parenthesis
Uncertain if intentional, spcomp doesn't warn upon use of ES function without parenthesis. Not sure if it would ever be useful in any manner to call function without parenthesis. Tested with `1.10.0-git6380` and `1.11.0-git6460`

```
#include <sourcemod>

enum struct Whatever {
	int x;
	char whatever[32];
	int test() {
		return sizeof Whatever::whatever;
	}
}

public void OnPluginStart() {
	RegAdminCmd("sm_test", cmdTest, ADMFLAG_ROOT);
}

public Action cmdTest(int client, int args) {
	Whatever w;
	ReplyToCommand(client, "%i %i", w.test, w.test()); // 0 32
	return Plugin_Handled;
}
```

Fyren:


compiler/semantics.cpp
```diff
@@ -1233,6 +1237,11 @@ FieldAccessExpr::AnalyzeEnumStructAccess(Type* type, symbol* root)
         return false;
     }
     if (field_->ident == iFUNCTN) {
+		if (!from_call) {
```
dvander: nit: tabs

Issue #426, opened 2019-11-30
Tag mismatch warning when comparing inheritance methodmaps


FortyTwoFortyTwo:
This compiles without warning in `1.10.0.6458`, but `1.11.0.6473` throws a tag mismatch.
Also worth noting that it seems to effect which way round into comparing variables, as comparing `INVALID_HANDLE` first over `array` compiles fine without warning
```js
public void OnPluginStart()
{
	ArrayList array = new ArrayList();
	
	if (array != INVALID_HANDLE)	// warning 213: tag mismatch
	{
		array.Push(1);
	}
	
	if (INVALID_HANDLE != array)	// no warning
	{
		array.Push(2);
	}
}
```

Fyren:
This actually has the same root issue in the new parser as in #424: 

https://github.com/alliedmodders/sourcepawn/blob/4b2eafae2a9db6e55f369513c588af3f11687e5c/compiler/expressions.cpp#L825-L827

The old parser swaps the operands around when the right side is a constant (like `INVALID_HANDLE`).  `matchtag` only walks up the methodmap chain on the right side, so when the new parser doesn't do this swap, we get a tag mismatch.

dvander:
Thanks, the fix for this is now in SourceMod.

PR #430, opened 2019-12-03
Fix commutative tag checks.

dvander:
Bug: issue #426
Test: compile-only/ok-compare-methodmap-constant

### DIFF
--- a/compiler/expressions.cpp
+++ b/compiler/expressions.cpp
@@ -349,17 +349,12 @@ matchobjecttags(Type* formal, Type* actual, int flags)
     int formaltag = formal->tagid();
     int actualtag = actual->tagid();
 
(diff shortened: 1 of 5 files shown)


Fyren:
This changes the behavior of `matchobjecttags`:

```
enum Handle { INVALID_HANDLE = 0 }
using __intrinsics__.Handle;

methodmap ArrayList < Handle { }
methodmap ArrayLister < ArrayList { }

public void main()
{
    ArrayList al;
    ArrayLister al2;

    if (al != INVALID_HANDLE) { } //1.11 mismatch
    if (INVALID_HANDLE != al) { }
    if (al != al2) { }
    if (al2 != al) { } //1.10, 1.11 mismatch
    if (main != INVALID_FUNCTION) { }
    if (INVALID_FUNCTION != main) { } //this PR error 132
}
```

Issue #412, opened 2019-11-09
Enum Struct const array bracket initialization


JoinedSenses:
Old enum structs allowed for const array bracket initialization. It works with new enum structs, however it gives warning 228. Arrayed enum structs don't work at all, giving error 001 and 010.

```
#include <sourcemod>
#pragma semicolon 1
#pragma newdecls required

enum struct Test {
        int i;
        float f;
        bool b;
        char str[32];
        float vec[3];
}

Test g_test[] = {
        /* warning 228: length of initializer exceeds size of the enum field
         * error 001: expected token: "}", but found "," */
        {1, 1.0, false, "value1", {1.0, 2.0, 3.0}},
        /* error 001: expected token: "-identifier-", but found "{" */
        {2, 2.0, true, "value2", {4.0, 5.0, 6.0}},
}; /* error 010: invalid function or declaration */

public void OnPluginStart() {
        /* warning 228: length of initializer exceeds size of the enum field */
        Test t = {1, 1.5, true, "value3", {7.0, 8.0, 9.0}};
        PrintToServer(
                "%i %0.1f %i %s %0.1f " // 1 1.0 0 value1 3.0
            ... "%i %0.1f %i %s %0.1f", // 1 1.5 1 value3 9.0
                g_test[0].i, g_test[0].f, g_test[0].b, g_test[0].str, g_test[0].vec[2],
                t.i, t.f, t.b, t.str, t.vec[2]
	);
}
```

dvander:
Thanks, the fix for this is now in SourceMod.

PR #431, opened 2019-12-03
Fix literal enum struct initializers.

dvander:
The root of this issue was incorrectly using the dim.array.length field
of enum field symbols. Additionally, this patch adds some new
restrictions. Braces are no longer allowed around string literals since
braces in initializers have semantic meaning.

Bug: issue #412
Test: tests/enum-structs/initializer


JoinedSenses:
thank you!

Issue #435, opened 2019-12-19
No tag mismatch warnings from switch cases


FortyTwoFortyTwo:
Compiled with `1.10.0.6458` and `1.11.0.6477`, `switch` does not have any checks when comparing different tags (`int`, `float`, `enum`, etc), not throwing any tag mismatch warnings. Unsure if intended or not, but this does not feel right.
```js
enum Foo
{
	Foo1,
	Foo2,
	Foo3
}

enum Bar
{
	Bar1,
	Bar2,
	Bar3
}

public void OnPluginStart()
{
	Foo whatever = Foo2;
	
	switch (whatever)
	{
		case Bar1: whatever = Foo2;
		case 3: whatever = Foo3;
		case 7.4: whatever = Foo1;
	}
}
```

Fyren:
It looks like there's no warning at last as far back as 1.6.3, but this seems like a reasonable feature request.

rumblefrog:
Since the PR was merged, this now can be closed I presume?

PR #436, opened 2019-12-20
Check switch cases for tag mismatches.

Fyren:
Bug: issue #435
Test: compile-only/warn-switch-tag-mismatch


dvander:
lgtm - thanks, this one has bugged me for a while.

PR #441, opened 2019-12-25
Set tag on array expressions

### LINKED ISSUE #288: Tag mismatch on float array literals passed to functions
This code:
```sourcepawn
public void OnPluginStart()
{
	PrintFloatArray( { 0.0, 1.0, 2.0, 3.0 }, 4 );
}

void PrintFloatArray( float[] arr, int count )
{
	for( int i = 0; i < count; i++ )
	{
		PrintToServer( "%f\n", arr[i] );
	}
}
```
results in a tag mismatch error.
`plugin.sp(3) : warning 213: tag mismatch`

The current hack people use to fix this is to just wrap the array in a `view_as<float>` like so: 
`view_as<float>({ 0.0, 1.0, 2.0, 3.0 })`

It'd be nice if the compiler was able to figure out the correct tag based on the contents to get the original code to work without warnings.

SlidyBat:
Array expressions would never get their tag set, which causes tag mismatches when the type of the array isn't an `int`.

Fixes #288

### DIFF
--- a/compiler/semantics.cpp
+++ b/compiler/semantics.cpp
@@ -908,6 +908,7 @@ ArrayExpr::Analyze()
 
     val_.ident = iARRAY;
     val_.constval = litidx - start_litidx;
+    val_.tag = lasttag;
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


dvander:
Can you add a test case for this?

dvander:
Thanks!

PR #445, opened 2020-01-07
Fix array able to be initialized with zero size

FortyTwoFortyTwo:
Array could not initialized with negative size or values exceeds size, but no errors thrown if size is zero. Even when error 9 states array size zero could not be made.

Also apparently `tf2_stocks.inc` have array based enum struct used for deprecated function. Since `TFResourceNames` array size is zero, build fails. Will make PR about that in sourcemod repo.


dvander:
Thanks!

PR #449, opened 2020-01-17
Begin moving parser.cpp to an AST.

dvander:
This patch series moves a bunch of top-level statement parsing code into the new AST-based parser. To preserve behavior, each statement is immediately resolved after being parsed. The major change here is that the parsing, name resolution, and evaluation now occurs in three distinct steps and is no longer intermingled.

This covers the easy global statement cases. The most difficult case is function statements, which cannot be moved until all non-global statements have been moved as well.

### DIFF
--- a/compiler/sctracker.h
+++ b/compiler/sctracker.h
@@ -3,6 +3,7 @@
 #define _INCLUDE_SOURCEPAWN_COMPILER_TRACKER_H_
 
 #include "lexer.h"
+#include "pool-allocator.h"
 #include "scvars.h"
 
 #define MEMUSE_STATIC 0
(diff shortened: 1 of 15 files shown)


compiler/sctracker.h
```diff
@@ -19,12 +20,10 @@ typedef struct funcarg_s {
 struct functag_t {
     functag_t()
      : ret_tag(0),
-       argcount(0),
        args()
     {}
     int ret_tag;
-    int argcount;
-    funcarg_t args[SP_MAX_EXEC_PARAMS];
+    PoolList<funcarg_t> args;
```
Fyren: With `-Wall`, gcc complains about signed/unsigned comparison because `argcount` was `int` but `PoolList::length()` is `size_t`.

PR #461, opened 2020-02-27
Include tag names in tag mismatch error message

asherkin:
I know this isn't quite perfect as the type resolution can be quite complex, and it is generally better to provide less information than bad information, but I think this improves DX in the vast majority of cases and is probably worthwhile doing unless there is anything super obviously wrong.

A lot of users have trouble understanding this error message, so a little bit of added context like this should go a long way. I've kept the message as "tag mismatch" instead of "type mismatch" for now, but that might be worth changing at some point as well. I've also changed "loose indentation" to clarify what it is checking a little better, as that is another common warning that people don't understand.


Fyren:
The loose indentation warning can get reported without mixing tabs and spaces.

dvander:
Thanks, this is a great idea - I've wanting this for a while but kept putting it off until the new parser is complete.

compiler/messages.h
```diff
@@ -259,11 +259,11 @@ static const char* warnmsg[] = {
     /*210*/ "possible use of symbol before initialization: \"%s\"\n",
     /*211*/ "possibly unintended assignment\n",
     /*212*/ "possibly unintended bitwise operation\n",
-    /*213*/ "tag mismatch\n",
+    /*213*/ "tag mismatch (expected \"%s\", got \"%s\")\n",
     /*214*/ "possibly a \"const\" array argument was intended: \"%s\"\n",
     /*215*/ "expression has no effect\n",
     /*216*/ "nested comment\n",
-    /*217*/ "loose indentation\n",
+    /*217*/ "mixed tabs and spaces used for indentation\n",
```
dvander: Indeed I think this will trigger if two statements at the same "level" have different indentation. Maybe: "inconsistent indentation (did you mix tabs and spaces?)"

We could always split it into two warnings if there's no good wording.

JoinedSenses: I like this idea.

asherkin: > Indeed I think this will trigger if two statements at the same "level" have different indentation. Maybe: "inconsistent indentation (did you mix tabs and spaces?)"

Yep, that's a lot better! Changed it to that.

asherkin:
> Thanks, this is a great idea - I've wanting this for a while but kept putting it off until the new parser is complete.

It was definitely a lot easier with the new parser changes - I did a first attempt at this a couple years back and there were a lot more changes required.

PR #463, opened 2020-03-08
Streamline arrays.

dvander:
This is a complete rewrite of the array parsing and emitting code, using
the new parser.

Array initializers are now parsed into the AST. This allows type
deduction, semantic checks, and code generation to occur in distinct,
separate phases. Because of this flexibility, and by virtue of not
having to maintain initials(), array internals have undergone
significant changes with this patch.

Indirection vectors are no longer offset-based. When dereferencing a
multi-dimensional array, each level points directly to the next level
without any offsetting required.

This makes array access much more efficient. For the following snippet:

	int[][] array = new int[l1][l2];
	return array[3][5];

We generate two fewer instructions, with the old listing on the left and
the new listing on the right:

	load.s.pri ffffffec     load.s.pri ffffffec
	add.c c                 add.c c
	move.alt                load.i
	load.i                  add.c 14
	add                     load.i
	add.c 14
	load.i

To keep local array declarations fast, arrays are now divided into three
contiguous segments:
  - The indirection vectors, or "iv".
  - Any prefilled data.
  - Zeroes, for any unfilled data

This structure exists for both local and global arrays. Global arrays
can be used as-is, since the IV section has absolute addresses. For
local arrays, a template structure is stored in the DATA section and
copied at runtime by a new INITARRAY opcode. INITARRAY replaces the
unused REBASE opcode. INITARRAY operates on uninitialized space in the
stack, by copying a source IV, rebasing its addresses to be stack-relative,
and then copying data and filling in zeroes.

Single dimension arrays go through INITARRAY, but specify an empty IV.
This allows the compiler to generate simpler code. Behind the scenes,
the JIT still performs a MOVS/FILL operation. Note that the old compiler
would emit a FILL and a MOVS, effectively writing to the array twice.
INITARRAY is more intelligent and only fills/copies as necessary.

Having all array initialization go through INITARRAY greatly simplifies
code generation, as does clearly separating the IV from array data. It
also eliminates the need for compile-time patching of indirection
vectors by rewriting the literal queue. The INITARRAY structure is much
easier to generate, since we can simply build the IV and data blocks
separately, in parallel, and then concatenate them together at the end.

An example of how code generation differs:

	char example1[] = "this is a test";
	int example2[10][20];

	Old:                    New:
	---------------------------------------------------
	stack -16               stack -16
	zero.pri                addr.pri example1
	addr.alt example1       initarray 0 0 4 0 0
	movs 10

	stack -840              stack -840
	zero.pri                addr.pri example2
	addr.alt example2       initarray 0x10 0xA 0 0xC8 0
	fill 0x348
	const.pri 0x10
	addr.alt example2
	movs 0x28

Note that for a normal array, the zero-fill operation is now implicit.
And for a multi-dimensional array, the manual template copy and fill
operations are now handled by the VM, which also eliminates a
double-write to the IV section.

The full format of initarray is:

	initarray <TemplateAddr> <IV Cell Count> <Data Copy Count>
	          <Data Fill Count> <Fill Value>

For supporting the deprecated "decl" keyword, the fill count is simply
left as 0, to avoid initializing the array.

Note that host applications that use multi-dimensional arrays must use
the new IPluginContext methods for reading arrays, since any hardcoded
methods based on offsetting will no longer work. For example, this
affects the sorting natives in SourceMod.

The length of arrays internally now represents their "true" length, not
their cell length. Conversion to cells happens when interacting with the
emitter. For example:

	char hello[] = "hello";

Previously this had a compile-time size of 8, because it requires 2
cells. The length is now correctly stored as 6 (5 characters plus a
terminator).

Lots of details around semantic checks of arrays have been tweaked. For
example, size overflow checks are more complete, postdims on arguments
now must have fixed sizes, and default array arguments are now type
checked.

There are two small syntax changes. The following is no longer legal:

	f(int a[]) {}

Whether or not it has a default initializer, postdims in transitional
syntax must have a fixed size. In this case no static size can be
determined. However, the following is now allowed:

	f(int[] a = {1, 2})

Also, sizeof() on a partially indexed array no longer returns the
computed array size (this number is entirely useless). Instead it now
returns the fixed size of that dimension.

Finally, "pragma dynamic" use cases should now be even rarer. The stack
size is automatically computed based on the stack frame of the largest
function.

There are many mechanical changes with this patch as well. The literal
queue has been removed, and replaced with a DataQueue class. The
DataQueue controls all accesses to the data section during code
generation, and allows faster movement of data into the queue.

The old array parsing and generation code, initials() and initvector(),
have been completely removed. These functions, as well as the ones to
update indirection vectors, were extraordinarily complex and
unmaintainable. They have been the source of many bugs and have been
poorly understood.

The declglb() and declloc() functions have been moved to the new parser,
removing a great deal of legacy code.

Because the AST-based code is so much more flexible, it is surprisingly
difficult to keep all of the old semantic restrictions in place.
Something like this:

	f(int[] a = new int[5])

Is now trivial to parse, and trivial to emit. However, this kind of
thing has been restricted until it can be better tested.


JoinedSenses:
Will the array default syntax support enum structs?

dvander:
Can you give an example of what you'd like to see work?

JoinedSenses:
Question was mostly out of curiosity, though I suppose the same way that you've allowed for enum structs to have bracket initialization in 1.11.

```
enum struct Thing {
  char[32] name;
  int value;
}

void f(Thing t = {"Doug", 69}) {
  // ...
}
```

dvander:
I think that will work after this patch. If not, it wouldn't be hard to get working.

dvander:
Another syntax trope that is no longer valid after this patch:

    bool BigArray[MAXPLAYERS + 1] = true;

dvander:
This pattern turns out to be very common, unfortunately - I'll special-case it and add a warning.

assyrianic:
Given this array rework. Will this change the restriction on how natives cannot return arrays?

dvander:
Probably there won't be any new features immediately available, but I want there to be something big in 1.12 - the breakage factor will be higher than usual (1.12 plugins won't run on older SM versions), so we need some kind of motivation to test it out. I don't know what that will be yet though, or how long it will take.

This PR does unblock two things: (1) moving forward with the parser rewrite, and (2) changing the representation of variables. (1) is the path toward returning arrays and having fully dynamic array allocation. (2) is the path toward new data types, like int64, proper array slices, etc.

dvander:
This is such a big patch, it's difficult to rebase and I'm going to commit it now. The final version has a very small impact on the forums corpus (less than 1% of plugins were affected). There's now a doc on how to fix the common errors in that 1%. The new code catches a lot of off-by-one bugs in string lengths which is nice to see.

Almost certainly more work will be needed on arrays, but this is a good start.

PR #481, opened 2020-05-13
Fix up location tracking for symbols

Fyren:
This could probably be used to report stuff like the location of a shadowed symbol instead of just the name.  Maybe other compiler messages could also use it.


dvander:
LGTM, thanks for doing this.

PR #491, opened 2020-05-26
Fix logical expression codegen and improve CommaExpr.

Fyren:


dvander:
Looks good, thanks for fixing this and for testing it so thoroughly.

Issue #484, opened 2020-05-16
[Feature Request] Overload operator ... for concatenation single chars.


TheByKotik:
Operator `...` already can concatenation `const char[]` type in compile time.
```
"Test"..."text."
```
But throw errors when trying concatenation single chars.
Sometimes requires define this as constant:
```

#pragma newdecls required
#pragma semicolon 1

enum // CSGO color codes.
{
	Color_Team = '\x03',	// Soft Violet, Blue, Yellow.
}

public void OnPluginStart ()
{
	PrintToChatAll( " \x01"...Color_Team..."Team color." );
	PrintToChatAll( 'T'...'E'...'X'...'T' );
}
```
Of course we can format it, it's low but unwanted CPU utilization.

PR #507, opened 2020-07-30
Add support for concatenating single chars onto strings #484

c0rp3n:
Partially addresses #484 though this does not support concatenating enums as this is not handled in the lexer like the current string concatenation is.
It would seem the string concatenation would have to be moved to a later stage of compilation so the value of the enum is know for that.
But this pull request does support char literals so defines could be used for the usage in #484.

Also added a test for this functionality `tests/basic/cat-str-char.sp` I've inspected the output of this and they are correct and haven't seen anyother side effects.


Fyren:
It would be helpful if you made the test check the output instead of being compile-only.  You just have to put the expected output in a file with the same name except with `out` as the extension.

c0rp3n:
Added a out file to the test and moved it to the correct folder, also realised that I type out the wrong char code so that got fixed too.

compiler/lexer.cpp
```diff
@@ -2187,24 +2187,38 @@ lex_string_literal(full_token_t* tok, cell* lexvalue)
...
 
         static char buffer[sLINEMAX + 1];
         char* cat = buffer;
-        while (*lptr != '\"' && *lptr != '\0' && (cat - buffer) < sLINEMAX) {
-            if (*lptr != '\a') { /* ignore '\a' (which was inserted at a line concatenation) */
-                *cat++ = *lptr;
-                if (*lptr == sc_ctrlchar && *(lptr + 1) != '\0')
-                    *cat++ = *++lptr; /* skip escape character plus the escaped character */
+        if (*lptr == '\"')
+        {
```
dvander, line 2195: Nit: brace on previous line:
` if (*lptr == '"') {`

compiler/lexer.cpp
```diff
@@ -2187,24 +2187,38 @@ lex_string_literal(full_token_t* tok, cell* lexvalue)
...
+                    *cat++ = *lptr;
+                    if (*lptr == sc_ctrlchar && *(lptr + 1) != '\0')
+                        *cat++ = *++lptr; /* skip escape character plus the escaped character */
+                }
+                lptr++;
             }
-            lptr++;
+        }
+        else
+        {
```
dvander, line 2207: Nit, collapse into one line:
`} else {`

c0rp3n:
Corrected the code styling / formatting to match the project.

dvander:
Thanks, looks good!

TheByKotik:
I just wanna reports it for future pr's.
```
	PrintToServer( '.'...'.' ); // Error.
	PrintToServer( '.'..."." ); // Error.
	PrintToServer( "."...'.' ); // Correct.
```

Issue #469, opened 2020-04-16
Function type incorrectly initialized to 0 on enum structs


Sikarii:
When initializing a variable of type `Function` in an enum struct, they seem to incorrectly default to 0.

I believe the Function type in this case **should** default to -1 (INVALID_FUNCTION) instead.

The following code results in:
- `Is null`: 1
- `Is INVALID_FUNCTION`: 0

```
enum struct MyStruct
{
    Function Handler;
}

public void OnPluginStart()
{
    MyStruct myStruct;

    PrintToServer("Is null: %d", (!!myStruct.Handler));
    PrintToServer("Is INVALID_FUNCTION: %d", (myStruct.Handler != INVALID_FUNCTION));
}
```

dvander:
Thanks for the report. This is an annoying case, because variables are supposed to be zero'd by default, but both 0 and -1 are invalid values for a function variable. It's way too late to fix the fact that we picked -1 for INVALID_FUNCTION.

I'm tempted to make the comparison implicitly check for both. Until I can finish the array refactoring PR, it's probably the easier route.

dvander:
This is now zero/null.

Issue #473, opened 2020-04-30
Newdecls required is not work for "any:..."


CrazyHackGUT:
```sourcepawn
#pragma newdecls required

stock void PriorityCenterTextAll(int iPriority = cellmin, char[] szFormat, any:...)
{
    char szBuffer[MAX_CENTER_TEXT];

    for (int i = 1; i <= MaxClients; i++)
    {
        if (IsClientInGame(i))
        {
            SetGlobalTransTarget(i);
            VFormat(szBuffer, sizeof(szBuffer), szFormat, 3);
            PriorityCenterText(i, iPriority, "%s", szBuffer);
        }
    }
}
```
This code compiles without any error about "old syntax". I think this is a bug, because on new syntax should be `any ...` instead `any:...`.

dvander:
Thanks, this has been fixed on master.

PR #520, opened 2020-09-04
Fix any: not triggering a newdecls warning.

dvander:
Bug: issue #469
Test: tests/compile-only/fail-old-any-decl.sp


SlidyBat:
I think this was meant to be linked to #473

dvander:
Yup, thanks.

PR #524, opened 2020-10-07
Fix enum increment/multiplier parsing

rtldg:
In Sourcemod 1.11 the following code fails to compile after moving enum parsing to new-parser.cpp in commit f598f1d22e9a596b70dd212ecf72d9655c965a9c
```
enum CheatType(+=1)
{
	Cheat_HighPerfects = 0,
	Cheat_Auto,
	Cheat_JumpSpam
};
public void OnPluginStart() { }
```
Errors:
```
% ./compile.sh test.sp

Compiling test.sp...
SourcePawn Compiler 1.11.0.6625
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2018 AlliedModders LLC

test.sp(1) : error 001: expected token: "{", but found "("
test.sp(6) : error 010: invalid function or declaration
test.sp(3) : warning 203: symbol is never used: "Cheat_JumpSpam"
test.sp(3) : warning 203: symbol is never used: "Cheat_Auto"
test.sp(3) : warning 203: symbol is never used: "Cheat_HighPerfects"
```

### DIFF
--- a/compiler/new-parser.cpp
+++ b/compiler/new-parser.cpp
@@ -212,8 +212,6 @@ Parser::parse_enum(int vclass)
 
     EnumDecl* decl = new EnumDecl(pos, vclass, label, name);
 
-    needtoken('{');
-
     cell increment = 1;
     cell multiplier = 1;
     if (matchtoken('(')) {
@@ -228,6 +226,8 @@ Parser::parse_enum(int vclass)
         }
         needtoken(')');
     }
+
+    needtoken('{');
 
     cell size;
     cell value = 0;


JoinedSenses:
The increment syntax was removed from sp by Bailopan because "nobody uses it". Your `(+=1)` is unecessary in this snippet as that is the default behavior, so you can remove it. if you need any other special incremental behavior, you'll have to use `#define` or manually write out the value in the enum.

JoinedSenses:
Hm. This whole time I thought increment syntax was intentionally removed behavior.

dvander:
I intended to remove it, but apparently broke it instead, and then assumed it had been removed. Whoops. As JoinedSenses pointed out, the "(+=1)" in the example is not necessary. Enums always increment by one.

I have a corpus of scripts (.sp forum attachments) and found this syntax to be completely unused except for the "+=1" case. If that's the only use case in the wild, it seems worthwhile to kill off the syntax completely.

rtldg:
I have no idea where people picked up this syntax and another place I've found the (useless) +=1 is [here](https://github.com/blankbhop/btimes/blob/c955961e938f9508c10957964d76af4a53d0fafa/addons/sourcemod/scripting/include/bTimes-replay3.inc#L18).
I figured it'd be easier to fix the compiler (since the code is still there) than to ensure people update the 1 or 2 addons that use this syntax whenever 1.11 drops.

JoinedSenses:
It's has been an obscure feature. I used it for bitshifted values to perform bitwise operations for flags.

dvander:
I'd be fine keeping it with a warning that the syntax will be removed in 1.12. It does pain me to break syntax without prior notice. In that case I'd take this PR and do the warning as a follow-up.

dvander:
The real problem is that documentation for obscure features is very poor, and we have to rely on de-facto behavior. But if no one uses a feature, it's like the riddle of a tree falling in the forest. No one's around to test it. So instead, we have to reverse engineer the functionality from the implementation, write a battery of tests around it to suss out the edge cases, then make sure the new functionality replicates that behavior, even if the actual functionality seems wrong in some cases. And guaranteed, we'd still somehow fail to replicate it.

So, the smart money is on removing or narrowing the syntax. For example, we could add a new syntax for declaring flags/bits, make the most notable use case more powerful, and drop the other vague/confusing ones.

JoinedSenses:
The code for increment syntax seems straight forward and non-invasive. Instead of removing a very simple syntax, and since as you say its undocumented, could we (I?) add a short section about it on a relevant wiki page and leave it in going forward rather than adding in some new stuff that might be even more confusing and just as obscure? I had wrongly made the assumption this behavior was already removed, so going forward after this PR - it might be more evident if this feature breaks for the people who use this feature.

dvander:
Okay, the other issue is I'm totally unconvinced it's a feature worth keeping or that people should be using. It exists in no other programming language. If the ability to create bit flags is something developers would find useful, let's just promote that to a first-class syntax instead.

JoinedSenses:
There was at least a few announcements for enum structs (which included some usage info), a section in the wiki for them, and people trying their darndest to break them (such as myself). They're at least more useful than the increment syntax as theyre one step closer to sorta pretending SP has objects. If you think itd be more valuable to provide a direct way of working with bit flags than #defines or enum increment syntax, then I am totally cool with that - just figured it would've been easier on you to leave in already functional behavior.

asherkin:
@dvander I think it'd be good to take this for 1.11 regardless of what happens to the syntax in the future - we'll need parsing working here to make it a nice user error one day anyway.

> There were at least a few announcements for enum structs (which included some usage info), a section in the wiki for them, and people trying their darndest to break them (such as myself).

dvander is referring "original" enum structs, i.e. tagged enum values, which has never been part of any SourcePawn documentation - not their new-syntax replacement (which exists because they were popular to use anyway).

asherkin (approved):

asherkin:
Thanks again for tracking down the bug here, I'm going to merge this even if we're not sure of the future of this syntax as a) the bug in the existing code is silly, b) the fix is right here ready to go, and c) 1.11 is getting close to going over the fence. It seems likely this syntax will still go away in the (relatively near) future, but hopefully properly with a deprecation cycle first - we don't need a repeat of the funcenum kerfuffle.

dvander:
This syntax will be removed in the next version. I'll adjust this code to emit a deprecation error/warning before the next release.

Adrianilloo:
I recently came to recompile a plugin I wrote about a year ago in SM 11 which uses few `+=2` incremented enums and found about the new error. Personally I think it's completely obvious that experienced people in SM or other languages find it useful.

naris:
Removing "enum multipliers" is a bad idea. Also, @dvander is completely wrong that this "exists in no other programming language" and @JoinedSenses is also wrong about it being an "obscure feature". enum multipliers come from c and pawn is a direct descendant of that in the form of small c.

This "obscure feature" is used a lot in any core that, especially in c and c++ such as OS kernel code (linux & windows), embedded systems, games and other code that relies on bitwise operations (such as the Source Engine this runs on)

Issue #525, opened 2020-10-13
"Not enough space on the heap" after #517


peace-maker:
After merging #517, calling any fake native messes up the `hp` pointer leading to error spam later.

https://github.com/alliedmodders/sourcepawn/blob/3c76ed00a3908868825e04da7e588e7b4595c752/vm/x86/jit_x86.cpp#L1403-L1421

Restoring the state after the fake native is misaligned when taking the slow path.

peace-maker:
Here's a short test case
```sourcepawn
#include <shell>

public main()
{
  printnum(dynamic_native(100));

  int[] dynarray = new int[13];
  dynarray[12] = 200;
  printnum(dynamic_native(dynarray[12]));
}
```

The argparser of spshell seems to be busted, which leads to not running tests using the JIT but only using the interpreter, but none of the other tests would have caught this either.

```
PS > spshell.exe --version
Usage: spshell.exe [-h] file
error: Missing value for 'file'.
PS > spshell.exe .\dynamic-native.smx --version
SourcePawn version: 1.11
API version: 20f/d
Just-in-time (JIT) compiler available.
```

PR #527, opened 2020-10-14
Quick (maybe) fix for dynamic natives.

Fyren:
Bug: #525 
Test: tests/basic/dyn-native.sp


peace-maker:
Works fine. Maybe add a test too?

Issue #526, opened 2020-10-13
Infinity compiling


Bara:
Found this on latest compiler versions (1.10.0.6497 and 1.11.0.6633).

Example:
```sp
#define TEST TEST

public void OnPluginStart()
{
    PrintToConsoleAll(TEST);
}
```

It compiles until I cancel the process.

And (same 1.11 compiler) with this code it also compiles until I cancel it:
```sp
#define TEST "Test"

public void OnPluginStart()
{
    PrintToConsoleAll(TEST);
}
```

Fyren:
The first example is a bug from upstream (that still exists there).  I can't reproduce any problem with the second.

C and C++ roughly disallow recursive macros, which I guess we can also do.  Or just count replacements and set a limit per line.

dvander:
(Keeping this open to change the way preprocessing replacement works.)

dvander:
Recursive macros are now supported and have the same behavior as C.

PR #530, opened 2020-10-17
[Lexer] Pre-processor maximum macro recursion depth

c0rp3n:
## Problem
As shown in #526 if you try to compile the below code, spcomp will enter an infinite loop, constantly trying to replace TEST with it self.
```
#define TEST TEST

TEST
```

## Solution
Add a maximum macro recursion check and error, this is currently arbitrarily set too 32, but could be increased.

This means spcomp shall now log a fatal error. "fatal error 197: maximum macro recursion depth reached"

### Edit

Should probably add a constant for the maximum depth and also print that in the error message, just need to decide upon a maximum depth first.


rumblefrog:
What's the rationale behind the fixed number 32?

c0rp3n:
No hard limit, just needs to have a limit to stop the infinite replacement 32 just seems an okay limit less it is just something like 1024.

dvander:
Having a preprocessor recursion depth limit seems like a good idea. But I'd argue the preprocessor is still broken. "#define TEST TEST" should not recurse at all.

PR #532, opened 2020-12-21
Add optional ';' after enum struct in spcomp

Wend4r:


Wend4r:
Explame:
```
enum struct Vector
{
    float X;
    float Y;
    float Z;
};    // <-
```

asherkin:
Thank you for the PR - a semicolon here is intentionally disallowed in order to provide consistency to the language, see #319.

This is indicated by the `require_newline(TerminatorPolicy::Newline);` call, just so you're aware, the correct change to make here to implement this would be to change the `TerminatorPolicy` to `NewlineOrSemicolon`.

PR #533, opened 2020-12-23
Implement perf JIT reporting APIs

asherkin:
Implemented here is the [original perf JIT format](https://github.com/torvalds/linux/blob/614cb5894306cfa2c7d9b6168182876ff5948735/tools/perf/Documentation/jit-interface.txt) which is widely supported but only supports simple name mapping, and the [newer "jitdump" format](https://github.com/torvalds/linux/blob/614cb5894306cfa2c7d9b6168182876ff5948735/tools/perf/Documentation/jitdump-specification.txt) which additionally supports embedding the machine code and source mapping information (along with a bunch of other things we don't currently have in the JIT).

Example usage:
```
# perf record -F 999 -g ./vm/spshell/linux-x86/spshell ./fib.smx
fib(42) = 267914296
[ perf record: Woken up 9 times to write data ]
[ perf record: Captured and wrote 2.124 MB perf.data (6339 samples) ]

# perf inject --jit -i perf.data -o perf.jit.data

# perf report -i perf.jit.data --stdio -q | head -n 15
    99.97%     0.00%  spshell  libc-2.29.so                [.] __libc_start_main
            |
            ---__libc_start_main
               main
               Execute
               sp::ScriptedInvoker::Invoke
               sp::PluginContext::Invoke
               sp::Environment::Invoke
               <jit invoke stub>
               fib.smx::main
               |
                --99.95%--fib.smx::fib
                          fib.smx::fib
                          fib.smx::fib
                          fib.smx::fib
```

Using the jitdump source mapping will probably require compiling perf from source, due to [this bug](https://lkml.org/lkml/2020/5/28/56), but it works [very well](https://gist.github.com/asherkin/3cc5a239cf77e0925baa33589bb16493).

### DIFF
--- a/vm/api.cpp
+++ b/vm/api.cpp
@@ -287,7 +287,7 @@ SourcePawnEngine2::LoadBinaryFromFile(const char* file, char* error, size_t maxl
     }
   }
 
-  if (!pRuntime->Name())
+  if (*pRuntime->Name() == '\0')
     pRuntime->SetNames(file, file);
 
   return pRuntime;
--- a/vm/jit.cpp
+++ b/vm/jit.cpp
@@ -25,6 +25,7 @@
 #include "plugin-runtime.h"
 #include "stack-frames.h"
 #include "watchdog_timer.h"
+#include "debug-metadata.h"
 #if defined(KE_ARCH_X86)
 # include "x86/jit_x86.h"
 #endif
@@ -79,14 +80,24 @@ CompilerBase::emit()
   pcode_start_ = method_info_->pcode_offset();
   code_start_ = reinterpret_cast<const cell_t*>(rt_->code().bytes + pcode_start_);
 
+  const char* function_name = rt_->image()->LookupFunction(pcode_start_);
+
+  debug_name_ = std::string(rt_->Name()) + "::" + function_name;
+
 #if defined JIT_SPEW
   Environment::get()->debugger()->OnDebugSpew(
-      "Compiling function %s::%s\n",
-      rt_->Name(),
-      rt_->image()->LookupFunction(pcode_start_));
+      "Compiling function %s\n",
+      debug_name_.c_str());
 
   SpewOpcode(stdout, rt_, code_start_, reader.cip());
 #endif
+
+  CodeDebugMap debug_map;
+
+  // DWARF has special tags for marking the prologue/epilogue, but they're not exposed
+  // by the jitdump format. As that information is useful for humans, we emit a couple
+  // of fake source file mappings to frame the actual function body.
+  debug_map.push_back({ masm.pc(), "<prologue>", 0 });
 
   emitPrologue();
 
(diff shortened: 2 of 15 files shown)


vm/jit.cpp
```diff
@@ -79,15 +79,31 @@ CompilerBase::emit()
   pcode_start_ = method_info_->pcode_offset();
   code_start_ = reinterpret_cast<const cell_t*>(rt_->code().bytes + pcode_start_);
 
+  const char* function_name = rt_->image()->LookupFunction(pcode_start_);
+
+  size_t rt_len = strlen(rt_->Name());
+  size_t len = rt_len + strlen("::") + strlen(function_name);
+
+  debug_name_ = std::make_unique<char[]>(len + 1);
```
dvander, line 87: Feel free to just use std::string in new code.

vm/linking.cpp
```diff
@@ -15,16 +15,277 @@
...
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+// __rdtsc
+#include <x86intrin.h>
+
+class PerfJitFile {
+ public:
+  PerfJitFile(bool self_delete) {
```
dvander, line 40: uber-nit: "explicit PerfJitFile(..."

vm/linking.cpp
```diff
@@ -15,16 +15,277 @@
...
+  //
+  // * GDB JIT Interface
+  //   https://sourceware.org/gdb/current/onlinedocs/gdb/JIT-Interface.html
+  //   Lets GDB show JIT frames when debugging, with source info.
+  //   Requires generating full ELF + DWARF objects in memory.
+
+#if defined(KE_LINUX)
+  // These have to be shared among all environments, as the filenames are only distinguished by PID.
+  // Once we support multiple environments per process we'll need some internal locking.
+  static std::unique_ptr<PerfJitFile> perf_jit_file = {};
```
dvander, line 268: It'd be better to find a globally visible place to put these, but for now this is fine. You shouldn't need the initializer though.

dvander (approved):

vm/linking.cpp
asherkin: I can't find a good spot elsewhere for these - any suggestions? I don't think we have any naturally global objects (Environments are the top-level instance), so they'd need to be global vars or static members of something (probably Environment?)

vm/jitdump.h
```diff
@@ -0,0 +1,66 @@
+// Below structs cribbed from tools/perf/util/jitdump.h in the Linux source tree.
```
dvander, line 1: Do we need any licensing stuff in this file?

asherkin: The header describes itself as GPLv2-only, so wouldn't be compliant/correct if we were to assert that here. The general consensus ([FSF](http://lkml.iu.edu/hypermail/linux/kernel/0301.1/0362.html), current state of Oracle v Google) is that pure data structs like this aren't copyrightable though, so I think just attribution like this is the right option.

vm/linking.cpp
```diff
@@ -15,16 +15,277 @@
...
+  // * GDB JIT Interface
+  //   https://sourceware.org/gdb/current/onlinedocs/gdb/JIT-Interface.html
+  //   Lets GDB show JIT frames when debugging, with source info.
+  //   Requires generating full ELF + DWARF objects in memory.
+
+#if defined(KE_LINUX)
+  // These have to be shared among all environments, as the filenames are only distinguished by PID.
+  // Once we support multiple environments per process we'll need some internal locking.
+  static std::unique_ptr<PerfJitFile> perf_jit_file;
+  static std::unique_ptr<PerfJitdumpFile> perf_jitdump_file;
```
dvander, line 269: It's fine here. My thinking re: moving toward Environment was, that's all code we know has to be audited for a multi-threaded refactoring. But squirreled away here, no one is going to find it.

So if we keep it here, can we put a static std::mutex in and hold the lock while writing the perf data? Then we know it's at least future-proofed.

vm/jitdump.h
dvander: ACK

vm/linking.cpp
asherkin: I moved the implementation out of `linking` and moved ownership of the files to Environment, I think that is looking a lot cleaner now indeed.

dvander (approved):
Thanks, this is indeed much nicer.

vm/api.cpp
```diff
@@ -287,7 +287,7 @@ SourcePawnEngine2::LoadBinaryFromFile(const char* file, char* error, size_t maxl
     }
   }
 
-  if (!pRuntime->Name())
```
KyleSanderson, line 290: This thing bugs me still, they're very different checks.

asherkin: The old one was broken by previous refactoring elsewhere, the new one is the intended behaviour.

PR #534, opened 2020-12-26
Hook Entity Output changes

Wend4r:
Fix return values type of natives to `bool` instead of `void`.
Added cases with outputs from attacker to `tests`.

Bundle with Pull Request in SourceMod branch.


asherkin:
These include files are copies of the SourceMod ones used for running compile tests - you want to make this PR against the SourceMod repo.

Issue #536, opened 2021-01-04
error 132: cannot coerce non-object type ... to object type null_t


schmidt39:
```
public void OnPluginStart()
{
    bool b;
    
    // ok
    ArrayList x = b ? new ArrayList(1) : null;
    
    // error 132: cannot coerce non-object type ArrayList to object type null_t
    ArrayList y = b ? null : new ArrayList(1);
    
    if (x || y) {}
}
```

sm 1.10

schmidt39:
`public void OnPluginStart()
{
	ArrayList a;
	int i;
	any x = i ? i : a; // warning 213: tag mismatch
	if (x) {}
}`

Fyren:
That warning is because the second and third operands there don't match.  The assignment part is irrelevant.

PR #537, opened 2021-01-04
Fix ternary rejecting expr ? null : obj

Fyren:
Bug: #536
Test: tests/compile-only/ok-ternary-obj-types.sp

`matchtag(some_methodmap, null_t)` works but `matchtag(null_t, some_methodmap)`.  Use `matchtag_commutative` instead so both `expr ? obj : null` and `expr ? null : obj` work.

### DIFF
--- a/compiler/semantics.cpp
+++ b/compiler/semantics.cpp
@@ -832,7 +832,7 @@ TernaryExpr::Analyze()
         return false;
     }
 
-    if (!matchtag(left.tag, right.tag, FALSE))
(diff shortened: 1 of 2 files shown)


c0rp3n:
The test would have better coverage if it isn't just compile only

Could just add prints for if they are null or not, but shall check the correct code is generated too, not just that it now compiles

Fyren:
Little slow on the response, but the test is only meant to cover the change made.

If the ternary operator in general is missing coverage, you can file an issue (or a PR).

PR #539, opened 2021-01-06
Add debug symbol interface

peace-maker:
Add `IPluginDebugInfo::CreateSymbolIterator` to iterate over all visible symbols at a given code address. Add new interfaces to expose the debug symbols `IDebugSymbol` and their type `IDebugType` as parsed from the `.dbg.*` and `rtti.*` sections in the .smx.
`IPluginDebugInfo` also received new functions to list all available functions in a plugin and in which source file they are defined in.

The VM now parses the RTTI from the .smx and validates the sections when they are present. Debug sections are only parsed and validated if `ISourcePawnEnvironment::EnableDebugBreak` was called and thus line debugging is enabled in the VM.
There is a cheap attempt to get some information from the old pre-rtti `.dbg.symbols` section in plugins compiled with pre 1.10 compilers, but the best result is achieved when used in combination with a recent compiler.

The `IDebugType` interface exposes multiple functions to check the type the object represents like `isInt32`, `isEnumStruct`, `isBoolean`. This allows for clean code when checking for a few types, but maybe an enum to use a `switch` on is good too?

This code has been cooking for years now and is the next iteration of #216. This allows to inspect the plugin state during runtime for debugging from an extension.


peace-maker:
Well, lets split this down.

PR #540, opened 2021-01-15
Added missing knife enums

nickdnk:
Hello

I'm having trouble figuring out if this belongs here or in the SourceMod repo, and I can see that some of these knife-weapon-PRs have been either rejected or merged, yet I don't see these changes on the master branch, so excuse me if this is not what you want, but it seems weird to me that this enumeration include is incomplete.


asherkin:
These are just old copies of the SourceMod files used for running unit tests. Unfortunately we can't move PRs between repositories.

These knives were added to SourceMod over a year ago, in PR alliedmodders/sourcemod#1126.

nickdnk:
Aha. That makes sense. It's still missing from here, however :https://sm.alliedmods.net/new-api/cstrike/__raw - I have close to zero knowledge about how you release those files on the website, but it took me for a spin that IDs came out of the game that I couldn't find anywhere.

nickdnk:
Or this, for that matter: https://sm.alliedmods.net/new-api/cstrike/CSWeaponID

asherkin:
The API documentation site doesn't get updated very often, it has a complex manual process.

You may want to try https://sourcemod.dev/, but the include files themselves are the only true reference.

nickdnk:
Okay. Thanks.

PR #541, opened 2021-01-22
Fix section header strings validation

vanz666:
According to the description, last_header_string (the nameoffs field of one of the sp_file_section_t structure) is an offset relative to header_strings_, instead of being an offset relative to the beginning of the buffer.

### DIFF
--- a/vm/smx-v1-image.cpp
+++ b/vm/smx-v1-image.cpp
@@ -135,7 +135,7 @@ SmxV1Image::validate()
 
   // Validate sanity of section header strings.
   bool found_terminator = false;
-  for (const uint8_t* iter = header_strings_ + last_header_string;
+  for (const char* iter = header_strings_ + last_header_string;
        iter < buffer() + hdr_->dataoffs;
        iter++)
   {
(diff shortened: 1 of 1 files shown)


asherkin:
Please could you attach an example (or include an explanation) of a broken plugin that this fixes.

The change looks good, but the written justification for it is based on non-code that could also easily be wrong, which makes it harder than it needs to be to validate this is correct.

vanz666:
> Please could you attach an example (or include an explanation) of a broken plugin that this fixes.
> 
> The change looks good, but the written justification for it is based on non-code that could also easily be wrong, which makes it harder than it needs to be to validate this is correct.

The meaning of this check is to ensure that the last section name string does not enter the area of ​​the decompressed region. 

Using the adminhelp.smx plugin as an example, you can see that stringtab = 0xFC, last_header_string (.dbg.locals) = 0xC0, here we start searching for a null character not even in the string table and the search can end at any first zero byte. 

Suppose that this check was successful and the last string in the string table does not end with a null character, then the search for a section by its name can read characters outside the string table (which is pointless).

[adminhelp.zip](https://github.com/alliedmodders/sourcepawn/files/5883326/adminhelp.zip)

peace-maker (approved):

peace-maker:
Whops, meant to squash, sorry

PR #551, opened 2021-03-22
Fix crash on undeclared static methodmap error

Headline:
Fixes https://github.com/alliedmodders/sourcemod/issues/1453

This patch just uses the name we looked up instead of the name from the (null) `field_` var.

Now errors with

```
SourcePawn Compiler 1.11
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2018 AlliedModders LLC 
test.sp(8) : error 105: cannot find method or property TestA.UndeclaredStaticFunc 
```

### DIFF
--- a/compiler/messages.h
+++ b/compiler/messages.h
@@ -129,7 +129,7 @@ static const char* errmsg[] = {
     /*102*/ "cannot find %s %s\n",
     /*103*/ "%s was already defined on this %s\n",
     /*104*/ "cannot find any methods for %s\n",
-    /*105*/ "cannot find method or property %s.%s\n",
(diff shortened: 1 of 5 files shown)


Headline:
A concern I have with this error is the fact that the symbol name isn't in quotes like other errors we have. What do ya'll think?

asherkin:
> A concern I have with this error is the fact that the symbol name isn't in quotes like other errors we have. What do ya'll think?

Probably worth adding the quotes for consistency.

dvander (approved):
The error messages aren't known for being consistent in any way, but probably quotes are the right call. Approved w/ or w/out though, since this was pre-existing.

PR #562, opened 2021-05-04
Start using Github Actions for CI

peace-maker:
Builds and tests on Linux, Windows and MacOS for x86 and x64.

The macOS builders don't support x86 builds anymore, so only test x64.
Build for arm and arm64 on Windows as well, but can't run the binaries yet.


dvander (approved):
This is great, thanks!

PR #568, opened 2021-06-20
Improve macro recursion support.

dvander:
The previous fix for this was intended to be a depth test, but actually
limited the number of macros per line, since substitution is iterative
and not recursive.

Instead, we now allow substituting macros infinitely. However each
substitution begins a "sequence", and this sequence is extended as new
substitutions are made. Any macro used inside this sequence is banned
from being substituted again. Once there are no more substitutions in
the sequence, a new one can begin, and the banned set is cleared.

This means that "#define TEST TEST" will now insert the characters
in "TEST", rather than infinitely perform substitutions, or return a
fatal error. This matches the C preprocessor.

Bug: issue #526
Test: ok-recursive-macro.sp


compiler/lexer.cpp
```diff
@@ -1379,9 +1381,22 @@ substallpatterns(unsigned char* line, int buffersize)
...
     int prefixlen;
 
-    int depth = 0;
+    std::unordered_set<std::string> blocked_macros;
+
+    auto enter_macro = [&](const char* start, size_t prefixlen, macro_t* out) -> bool {
+        if (!find_subst(start, prefixlen, out))
+            return false;
+        if (blocked_macros.count(out->first)) {
+            printf("%s\n", std::string((const char *)start, prefixlen).c_str());
```
Fyren, line 1390: Left this debug print in.

Also, this PR is against the other PR branch and not master.

dvander: Thanks, fixed.

Issue #245, opened 2018-10-01
Including files from withing sub directory with relative path is inconsistent/fails


Impact123:
## Setup
I have a directory layout like this.

```
│   plugin.sp
└───includes
    │   extension.inc
    └───extension
            file.inc
```

`plugin.sp` includes `extension.inc` like this
```SourcePawn
#include "includes/extension"
```

`extension.inc` includes `file.inc` like this
```SourcePawn
#include "extension/file"
```

spcomp is in the `PATH` and i compile from the directory where `plugin.sp` is like this
```bash
spcomp plugin.sp
```
A test case that compiles on linux but fails on windows can be found here: [includebug.zip](https://github.com/alliedmodders/sourcepawn/files/2432797/includebug.zip).

The version i used were
**Windows**: `1.10.0.6336`
**Linux**: `1.10.0.6335`

## Description
The error i get is this
```
SourcePawn Compiler 1.10.0.6336
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2018 AlliedModders LLC

includes/extension.inc(1) : fatal error 183: cannot read from file: "extension/file"

Compilation aborted.
1 Error.
```

If i change the include inside `extension.inc` to
```SourcePawn
#include "includes/extension/file"
```
it works.

The problem is that this is inconsistent between windows and linux. If i make the change described here then it works on windows but not on linux.

## Expected solution
Ideally one would expect that the include paths are always relative to the file that includes them 
like in the attached [includebug.zip](https://github.com/alliedmodders/sourcepawn/files/2432797/includebug.zip).

Please let me know if you need more information.

dordnung:
The problem is on this line: https://github.com/alliedmodders/sourcepawn/blob/master/compiler/lexer.cpp#L220
It tries to search relative to the last include file (which is **includes/extension** in the given example). But on the given line it retrieves the base path by finding the last directory separator. But there is no windows directory separator in **includes/extension**, which is the reason it works on linux but not on windows systems

Impact123:
I don't feel confident enough in this to create a pull request, but here's some proof of concept patch that appears to fix this. I have only briefly tested this on windows
```diff
diff --git a/compiler/lexer.cpp b/compiler/lexer.cpp
index 28a5457..205b892 100644
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -205,7 +205,16 @@ doinclude(int silent)
 
     i = 0;
[...]

PR #570, opened 2021-06-24
Sync include behavior for subdirectories between windows and linux

Pheubel:
This PR aims to solve https://github.com/alliedmodders/sourcepawn/issues/245 (Including files from withing sub directory with relative path is inconsistent/fails)

Motivation:
In larger plugins subdirectories are used to create structure in the project, however with the current handling of include paths it would make cross platform development and automation of multi platform CI/CD a pain.

Potential issues:
This change would be breaking for current projects developed on windows with a subdirectory structure. It would be relatively easy to fix for project maintainers by making the includes relative to the file that is including them.

Credits go to @impact123 for coming up with the proposed solution.

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -210,9 +210,10 @@ doinclude(int silent)
         if (DIRSEP_CHAR != '/' && *lptr == '/') {
             name[i++] = DIRSEP_CHAR;
             lptr++;
-            continue;
         }
-        name[i++] = *lptr++;
+        else {
+            name[i++] = *lptr++;
+        }
     }
     while (i > 0 && name[i - 1] <= ' ')
         i--; /* strip trailing whitespace */
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -206,7 +206,14 @@ doinclude(int silent)
 
     i = 0;
     while (*lptr != c && *lptr != '\0' && i < sizeof name - 1) /* find the end of the string */
+    {
+        if (DIRSEP_CHAR == '\\' && *lptr == '/') {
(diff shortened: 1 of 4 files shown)


dvander:
Can you clarify what is going on here? Why would an include path ever include a backslash?

Pheubel:
Sorry for the late response.

in [osdefs.h](https://github.com/alliedmodders/sourcepawn/blob/master/compiler/osdefs.h#L69) the directory separator for windows is set to `\`. In the plungefile function it uses this directory separator to determine the relative path.

If we take impact's example on a windows environment:
```
│   plugin.sp
└───includes
    │   extension.inc
    └───extension
            file.inc
```
where `plugins.sp` includes `extension.inc` and `extension.inc` includes `file.inc`, for the directory separator `/` is used to separate directories.

For simplicity sake, we assume that plugin.sp is the root of the path, since it is the file being compiled. 
When `plugin.sp` tries to include `extension.inc` via `includes/extension.inc`, the `inpfname` reference currently holds `\plugin.sp` it will include the file just fine and read `include/extension.inc`.
When `extension.inc` tries to include `files.inc` via `extension.inc`, the `inpfname` reference currently holds `\include/extension.inc`. Since it is searching for `\` to determine the relative path it will look on the same directory level as when compiling has started.

by changing the `/` occurrences to the directory separator for the platform the `inpfname` refernce will hold `\include\extension.inc` and plungefile will see `\include\` as the current path.

I hope i did a decent job at trying to explain it, it was a lot of magic boogaloo that i had to stare at for a while

Fyren:
Thanks for the PR.

PR #574, opened 2021-06-29
Add a new static_assert helper.

dvander:
The preprocessor directive `#assert` can only reference preprocessor
constructs. There is no way to assert compile-time constants defined in
SourcePawn.

This change introduces the `static_assert` statement, which has two
forms:

	static_assert(expression)
	static_assert(expression, "message")

The expression must be constant, eg, function calls are not allowed, and
any identifiers must be enum identifiers or constant variables. The
message is optional, but if provided, it must be a string literal.
Variables are not allowed.

If the expression evaluates to zero or false, an error is raised and
compilation will fail.

`static_assert` is only valid at global scope.

Grammar:

	static-assert ::= "static_assert" "(" const-expr ("," string-literal)? ")"

There are also a handful of minor changes in this series. A warning is now raised when preprocessor directives reference language constructs. This will be come an error in the following release.

Finally, a warning is raised that enum multipliers are slated for removal as well.


Fyren:
Not a big deal, but this only asserts and doesn't warn:
```
const int a = 1;
#define b 1
#define c 2

#assert a == b
#assert b == c

public void main() {}
```

Without the second assert, the warning is displayed for the first.  Is it getting ignored first pass and just hitting the second assert?

dvander:
Thanks for finding this, #assert actually triggers a fatal error which setjmp's back to main. So it's completely bypassing the second pass. I'm not sure how to fix it. I could force the warning to always emit in the first pass I guess, and then ignore it in the second pass.

PR #581, opened 2021-07-05
Fix crashes when SMX plugin doesn't have a function name

### LINKED ISSUE #580: Server crashes when the SMX plugin doesn't have a function name
After the changes in #533 , my server crashes when a SMX plugin doesn't have function name.

In line 85, the null `function_name` is dereferenced.
https://github.com/alliedmodders/sourcepawn/blob/07f9c462ab474a7a0a649e40c404de3da5cb32b2/vm/jit.cpp#L83-L85

At a lower level, it looks like this
image
image

Wend4r:
Fixes #580
Past changes #533

### DIFF
--- a/vm/jit.cpp
+++ b/vm/jit.cpp
@@ -82,7 +82,7 @@ CompilerBase::emit()
 
   const char* function_name = rt_->image()->LookupFunction(pcode_start_);
 
-  debug_name_ = std::string(rt_->Name()) + "::" + function_name;
+  debug_name_ = std::string(rt_->Name()) + "::" + (function_name ? function_name : "<unknown function>");
 
 #if defined JIT_SPEW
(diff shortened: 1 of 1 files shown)


Wend4r:
I checked, this PR to fix the crash problem. The server is working again

dvander (commented):
Functions must be named. This should fail validation.

Wend4r:
> Functions must be named. This should fail validation.

Then does it make sense to return not `nullptr`, but `""` or `"<unnamed function>"` in the SMX image
https://github.com/alliedmodders/sourcepawn/blob/master/vm/smx-v1-image.cpp#L719-L739

Fyren:
Validation would mean a check to make sure there is a name.  Failing validation means the plugin will error and not load.

We are not going to support stripping out debug/RTTI info.

PR #584, opened 2021-07-08
Dry run opt

Deathreus:


peace-maker:
Why? I guess we need an issue template.

asherkin:
> Why?

The main use case here is syntax checking for editor integration - because spcomp always appends `.smx` to the output filename you can't just output to `/dev/null` (or platform equiv), so you might as well just avoid the output stage completely (which does take some time).

dvander:
"Dry run" implies something destructive or irreversible would otherwise take place. I'd prefer this to be called "syntax only" mode (with the argument being --syntax-only). That or change the behavior of `-o` (or add a `--no-extension` flag).

I didn't see a justification for the `stderr` change in the commit message. It's better to put that behind a flag (eg `--use-stderr`) because there's a lot of potential for breakage there.

Deathreus:
> I didn't see a justification for the `stderr` change in the commit message. It's better to put that behind a flag (eg `--use-stderr`) because there's a lot of potential for breakage there.
That one wasn't supposed to be included in the PR, but making it a flag is a good idea for the people that want/need the option (I did for my use-case)

> "Dry run" implies something destructive or irreversible would otherwise take place.
Not sure how that gets implied, the literal definition of "dry run" is a test run before the real run

dvander:
"A dry run (or a practice run[1]) is a testing process where the effects of a possible failure are intentionally mitigated."

There are no failure effects for a failed run of spcomp.

sapphonie:
"dry-run" is very commonly used to mean "no output file" lol

dvander:
I don't really want to bikeshed this. `--syntax-only` is what will move the PR forward. 

Did line endings change in scvars.h or something?

Deathreus:
> Did line endings change in scvars.h or something?

I think the browser merge conflict editor did that, I forgot my sourcepawn was out of date and didn't want to add a new commit just pulling upstream(as I am too lazy to rebase)

dvander:
Thanks - I've manually rebased + merged this with a few tweaks:

 - Fixed line endings in scvars.h
 - Dropped -u and -S short forms for the new options. I'm leaning against short form arguments in the future. As we run out of letters they become difficult to remember.

Deathreus:
> * Dropped -u and -S short forms for the new options. I'm leaning against short form arguments in the future. As we run out of letters they become difficult to remember.

Gotcha

Issue #182, opened 2018-02-18
Compiler define __FILE__


CrazyHackGUT:
What about a implementing define with name `__FILE__` which contains current included filename?

Fyren:
We'd accept that if you're willing to make another PR, especially since `__LINE__` is pretty useless with multiple files without it.

CrazyHackGUT:
I tried implement this is manually before creating issue.
Perhaps I understood [how to create string constants](https://github.com/alliedmodders/sourcepawn/blob/master/compiler/sc1.cpp#L568), but the method is not perfect. Perhaps I found an [input point](https://github.com/alliedmodders/sourcepawn/blob/master/compiler/sc2.cpp#L262) for reading all the included files.

But something went wrong...
Constantly, I get the `core` in the value of the current file. Even in another included file.
![image](https://i.imgur.com/0kT4O9B.png)

What am I doing wrong?
I create a constant once, and while reading any file, I just update its value.

Fyren:
I took a quick look at `insert_subst`, and I think it makes a copy of the given string (`insert_subst` -> `insert_stringpair` -> `duplicatestring`).  Is there something wrong with `add_constant` with `inpfname`?

CrazyHackGUT:
`inpfname` always equals plugin original source. If i try compile `something.sp`, include in `something.sp` another different file like `something/constants.sp`, insert `__FILE__` in `constants.sp`, I whatever receive `something.sp`, but i need `something/constants.sp`.

dvander:
What does the patch look like?

CrazyHackGUT:
I'll post him when I came home.

Fyren:
> `inpfname` always equals plugin original source.

I don't think that's true.  `plungequalifiedfile` pushes `inpfname` to a stack, sets it to the new file, and `readline` pops it back out at EOF.

Fyren:
Implemented via #585.

PR #585, opened 2021-07-09
added __FILE_PATH__ and __FILE_NAME__ definitions

Pheubel:
This PR implements a feature requested in [#182 (Compiler define \_\_FILE\_\_)](https://github.com/alliedmodders/sourcepawn/issues/182)

It adds two new macros that can be used for debugging purposes, like `__LINE__`:
* ``__FILE_PATH__`` - A string containing the full path the the document where it is used in.
* ``__FILE_NAME__`` - A string containing just the name of the current file without a path.

I used ``__BINARY_PATH__`` and ``__BINARY_NAME__`` as examples on how to implement these definitions.

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -168,6 +168,11 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
             if(!pcwd) {
                 error(194, "can't get current working directory, either the internal buffer is too small or the working directory can't be determined.");
             }
+
+#if defined __MSDOS__ || defined __WIN32__ || defined _Windows
+            // make the drive letter on windows lower case to be in line with the rest of SP, as they have a small drive letter in the path
+            cwd[0] = tolower(cwd[0]);
+#endif
         }
     }
 
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -134,6 +134,8 @@ int
 plungefile(char* name, int try_currentpath, int try_includepaths)
 {
     int result = FALSE;
+    char* pcwd = NULL;
+    char cwd[_MAX_PATH];
 
     if (try_currentpath) {
         result = plungequalifiedfile(name);
@@ -153,6 +155,13 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
                 }
             }
         }
+        else {
+            pcwd = getcwd(cwd, sizeof(cwd));
+            error(224, pcwd);
+            if(pcwd == NULL){
+                error(1, "can't get current working directory, either too small or can't be determined.");
+            }
+        }
     }
 
     if (try_includepaths && name[0] != DIRSEP_CHAR) {
@@ -165,7 +174,16 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
         }
     }
 
-    set_file_defines(inpfname);
+    if(pcwd != NULL) {
+        char path[_MAX_PATH];
+        SafeSprintf(path, sizeof(path), "%s%s", pcwd, inpfname);
+        set_file_defines(path);
+        error(224,pcwd);
+        // set_file_defines(inpfname);
+    }
+    else {
+        set_file_defines(inpfname);
+    }
 
     return result;
 }
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -163,10 +163,9 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
                     result = plungequalifiedfile(path);
                 }
             }
-        }
-        else {
+        } else {
             pcwd = getcwd(cwd, sizeof(cwd));
-            if(pcwd == NULL) {
+            if(!pcwd) {
                 error(194, "can't get current working directory, either the internal buffer is too small or the working directory can't be determined.");
             }
         }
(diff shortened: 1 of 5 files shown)


Pheubel:
When a file from an included directory is included, it will use the name of the directory that is to be included. For example you marked `full/path/addons/sourcemod/scripting/include` to be included during compilation, using ``__FILE_PATH__`` in `tf2.inc` would return `include/tf2.inc`. If this is a problem i can try to look more into it.

compiler/lexer.cpp
```diff
@@ -163,9 +164,43 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
...
+            break;
+        }
+    }
+
+    if (inptr == NULL) {
+        inptr = file;
+    }
+
+    snprintf(newpath, sizeof(newpath), "\"%s\"", file);
+    snprintf(newname, sizeof(newname), "\"%s\"", inptr);
```
dvander, line 198: Do we need both of these? Seems like just the path would be sufficient. That would simplify this code a fair bit too.

compiler/lexer.cpp
```diff
@@ -58,6 +58,7 @@ static cell litchar(const unsigned char** lptr, int flags);
 
 static void substallpatterns(unsigned char* line, int buffersize);
 static int alpha(char c);
+static void set_file_defines(char* file);
```
dvander, line 61: spcomp is C++ now, and uses STL, so we can make this a bit prettier:

`static void set_file_defines(char* file);`

compiler/lexer.cpp
dvander: This becomes `file.c_str()` and `inptr.c_str()`

compiler/lexer.cpp
```diff
@@ -163,9 +164,43 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
...
+#if defined WIN32 || defined _WIN32
+            || file[i] == '\\'
+#endif
+        ) {
+            inptr = &file[i + 1];
+            break;
+        }
+    }
+
+    if (inptr == NULL) {
```
dvander, line 193: This becomes `if (inptr.empty()) {`

compiler/lexer.cpp
```diff
@@ -163,9 +164,43 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
...
+    for (i = len - 1; i < len; i--) {
+        if (file[i] == '/'
+#if defined WIN32 || defined _WIN32
+            || file[i] == '\\'
+#endif
+        ) {
+            inptr = &file[i + 1];
+            break;
+        }
+    }
```
dvander, line 191: This becomes something like:

```
std::string inptr;
for (size_t i = file.size() - 1; i < file.size(); i--) {
    if (condition) {
        inptr = file.substr(i);
        break;
```

Finally to fix the backslash bug mentioned offline, you could do something like:
```
auto pos = file.find('\\');
while (pos != std::string::npos) {
    file.insert(pos + 1, '\\');
    pos = file.find('\\', pos + 2);
}
```

(Completely untested!)

compiler/lexer.cpp
Pheubel: changed the signature to `static void set_file_defines(std::string file);`

compiler/lexer.cpp
```diff
@@ -152,6 +164,12 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
                 }
             }
         }
+        else {
+            pcwd = getcwd(cwd, sizeof(cwd));
+            if(pcwd == NULL) {
```
dvander, line 169: `if (!pwd) {`

compiler/lexer.cpp
```diff
@@ -163,9 +181,44 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
             result = plungequalifiedfile(path);
         }
     }
+
+    if(pcwd != NULL) {
```
dvander, line 185: `if (!pwd) {`

compiler/lexer.cpp
```diff
@@ -152,6 +164,12 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
                 }
             }
         }
+        else {
```
dvander, line 167: "else" should be on the same line as the previous brace, eg: `} else {`

compiler/lexer.cpp
```diff
@@ -163,9 +181,44 @@ plungefile(char* name, int try_currentpath, int try_includepaths)
             result = plungequalifiedfile(path);
         }
     }
+
+    if(pcwd != NULL) {
+        char path[_MAX_PATH];
+        SafeSprintf(path, sizeof(path), "%s%s", pcwd, inpfname);
+        set_file_defines(path);
+    }
+    else {
```
dvander, line 190: Same thing here, `} else {`

dvander (commented):
Looks good, some minor nits and still needs a unit test case - see tests/basic for examples and they're very easy to run (`python tests/runtests.py objdir`)

compiler/lexer.cpp
Fyren: I don't think the "if" and parens nuzzling was changed, but everything else looks good to me.  Thanks for doing this and adding tests.

Pheubel: i don't really get what you mean with that, i could be missing something but i think i got everything cleaned up.

Issue #501, opened 2020-07-15
Large array declarations has different behavior


kice:
```
#include <sourcemod>

public void OnPluginStart()
{
    int buf1[4096]; // Totoally A okay
    // int buf2[5120]; // [SM] Exception reported: Not enough space on the stack
    // int buf3[8192]; // [SM] Exception reported: Instruction contained invalid parameter
}
```

```
SourcePawn Compiler 1.11.0.6520
```
```
 SourceMod Version Information:
    SourceMod Version: 1.11.0.6520
    SourcePawn Engine: 1.11.0.6520, jit-x86 (build 1.11.0.6520)
    SourcePawn API: v1 = 5, v2 = 12
    Compiled on: Mar 11 2020 21:26:19
    Built from: https://github.com/alliedmodders/sourcemod/commit/ece44718
    Build ID: 6520:ece44718
    http://www.sourcemod.net/
```

I think for array size larger than stack size, it should report `Not enough space on the stack`. I assume very larger array size breaks the VM bytecode.

dvander:
I think the array rewrite PR fixes this. In the meantime, you can do this at the top of your plugin:

    #pragma dynamic 16384

Which will tell the VM to allocate a much bigger stack for your plugin.

kice:
Also I found that if I could declear the array as a global variable without using `#pragma dynamic`, then it looks like we don't have a limit in gloabl scope.

```
#include <sourcemod>

int buf[10485760];

public void OnPluginStart()
{}
```

Then I have three questions:

1. Is this usage intentional? Or `#pragma dynamic` only effect non-global scope?

2. And in this case, I decleared 40MB array (10485760 int), but the memory usage increased more than 120MB (by Windows TaskMgr commit size of srcds.exe), compare to before loading the plugin. 

3. Beside using ArrayList, do we have a better way to store large amount data out of the box?

dvander:
(1) Global scope is static, so computing the upper memory bound is trivial. Not so for function calls which can be recursive, so the compiler doesn't even try. The array rewrite PR fixes this by adding some heuristics. But to answer your question, #pragma dynamic only affects the memory pre-allocated for function calls.

[...]

PR #590, opened 2021-07-10
Adjust memory based on inferred maximum stack size.

dvander:
This eliminates a common use case for #pragma dynamic.

Bug: #501
Test: new runtime test


asherkin:
Did you already write this somewhere? I'm sure I've seen this code - was it part of the array rewrite branch?

dvander:
Yeah it's on the array branch, which is well into rebase-necessary territory.

PR #639, opened 2021-08-23
Add some general includes files

Bara:
for json it exists [this include](https://github.com/clugg/sm-json/tree/master/addons/sourcemod/scripting/include), but it's not compatible since RIPExt use the same name "json.inc".


dvander:
Thanks, I'll include these.

PR #644, opened 2021-08-24
Ensure function signatures are checked when performing casts.

dvander:
This makes it invalid to view_as to a different signature type. In
fact, view_as should *never* be used on functions. This is the first
step in battening down the hatches.


RusJJ:
Now everyone cannot "backup" a function using DataPack and later "unpack it" and use in a Query as a callback. What now?

PR #655, opened 2021-08-29
Eliminate reparsing.

dvander:
This is the biggest change to the compiler in SourcePawn history, and it was
only made possible by a long, slow crawl of incremental refactoring. It
will enable feature development in a way that simply wasn't possible
before.

Normal compilers have a pipeline of lexing, parsing, semantic analysis, and code
generation. Pawn interleaved this process without any sort of abstraction,
making the implementation completely unmaintainable and totally unamenable to
complex improvements.

There were many, many bad effects from this. Pawn had very little capability to
resolve cyclic references, a very common complaint with enum structs and
methodmaps. It had no way to resolve symbols that were used before definition.

To work around this, Pawn would repeat the entire compilation pipeline at least
twice. The first pass would attempt to collect symbolic information which then
got partially thrown away and retained into the next pass. Getting this to work
accurately was extremely error prone. The first pass aggressively errored
because it was incomplete, and these errors had to be ignored to make sure
symbolic information was populated for the next pass. Then second pass then had
to distinguish between an incorrect pre-population and a correct one. A third or
fourth pass was usually required due to the way operators were implemented. It
was a mess.

Over the past two years we carefully broke apart the parser block by block into
distinct phases. However we were unable to change the reparse mechanism until
the entire codebase had been converted. We have finally reached that point, and
thus, we can do the final refactoring that makes this a normal compiler.

Backstory out of the way, the first change is that the binding phase is split
into two sub-phases. The first phase is "name entering" which scans the parse
tree for global names and creates empty symbolic information. The next phase is
name resolution, which pairs each name use to its corresponding declaration.

Types are now determined in the name resolution phase. This is another big
change, and it brings us one step closer to replacing tags with types.

The parse tree is now evaluated in its entirety, rather than piece by piece.
This means that if the binding phase fails, compilation stops. We do not attempt
to analyze semantics if we can't resolve names. We do not attempt to resolve
names if there is a parsing error. A few tests had to be split up to account
for this.

All the code to deal with "stages" is gone. sc_status is gone. fetchfunc() is
gone. We no longer need to add member functions into the global symbol table.
Function overloading is now easy to implement. Operators now bind by type, not
by name. The "static" scope is now an actual scope.

Methodmaps and enum structs can now reference other methods on the same type.

Functions with errors will now always error. Stocks that happen to be
unused, but have syntactic errors, will now fail compilation.

And most important, reparsing is gone. The compiler is now between 2-3X
faster.


justkamiii:
Cheers! 🎉

dvander:
There is still a lot of cleanup to do here - for example, I want to unify the scope hierarchy (local, statics, and globals are not properly layered). I would also like to eliminate markusage() which is a crutch from the old reparsing model. The same goes for the referrer algorithm which finds dead functions (and is currently basically doing nothing).

But as-is this is a big and important step, so I'm checking it in now to avoid bitrot. When SourceMod branches to 1.12 I'll be able to integrate the new compiler & VM.

PR #657, opened 2021-09-06
Parse include directives as statements.

dvander:
This is a huge hack, but it will make it much easier to properly clean
up the implementation of scopes. The problem is that #include statements
can appear anywhere, eg:

     public void
     #include "blah.sp"
     () {}

In the current system, static variables are globals that are tied to a
specific file. This requires attaching a file indicator to every name
lookup. It would be much cleaner to place statics into a dedicated
scope.

In the above example, "blah.sp" introduces a new static scope while in
the middle of parsing a function. This is extremely difficult to deal
with and exposes a bunch of crazy edge cases.

The effect of this change is that #include is now legal ONLY as a
top-level statement. It acts much like a declaration now. Due to the
lexer being very tricky, this works by injecting temporary tokens into
the token stream. These tokens are used by the parser to open the file
and reset the lexer.

Thus the above code is now illegal, as are lots of other #include
patterns that would be legal in C.

### DIFF
--- a/compiler/errors.cpp
+++ b/compiler/errors.cpp
@@ -213,7 +213,7 @@ MessageBuilder::~MessageBuilder()
     ErrorReport report;
     report.number = number_;
     report.fileno = where_.file;
-    report.lineno = where_.line;
+    report.lineno = std::max(where_.line, 1);
     if (report.fileno >= 0) {
         report.filename = get_inputfile(report.fileno);
     } else {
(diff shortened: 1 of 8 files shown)


TheByKotik:
I need to comment it...

For now SP don't have native features as inline functions and creating template with function substitutions.
I'm active using this feature of #include directive in many plugins.

Perhaps now is too early to remove this feature?

dvander:
This is not a supported use case of #include, and using it to token paste is super unhygienic. I'd recommend factoring your code in a way that emphasizes small functions and if you need new high level language features please feel free to start a discussion.

PR #688, opened 2021-10-31
Rewrite heap tracking.

dvander:
Since the AMX Mod X days we've dealt with heap allocations by trying to
precisely free them, similar to stack opcodes. This gets complicated by
the ternary operator which could have different allocations on each
side. It's further complicated by the method verifier, not to mention
nuances in evaluation ordering.

This patch throws away the entire concept of the "heap tracker". Heap
allocations are no longer precisely tracked, and the TRACKER opcodes are
banished from new binaries.

The new scheme is to save and restore the heap pointer via two new
codes: OP_HEAP_SAVE and OP_HEAP_RESTORE. These opcodes are emitted with
a much simpler granularity than the old tracker logic.

If a node requires heap allocations, we mark the node, and then set a
"pending allocation" bit. The nearest block statement then grabs this
bit and assumes responsibility for emitting SAVE/RESTORE opcodes. This
is called a heap scope. Control-flow statements can have their body
create a heap scope (to avoid memory explosion in loops). The "advance"
expression of a "for" statement also creates a heap scope.

It is never necessary to create a heap scope, but the more granular they
are, the less risk there is of running out of memory. We've struck a
balance here that roughly matches how stack allocations work.

We do not create a heap scope for the function body. Instead,
PluginContext::Invoke always restores the heap pointer. It now also
restores the stack pointer as well, making it unnecessary to do any
extra code generation on return paths.

The unfortunate aspect of this patch is that the Semantics pass needs to
precisely mimic when CodeGeneration will allocate on the heap, so we can
build the heap scope boundaries. To future proof this, the code
generator asserts that heap allocations are tied to a node that was
properly marked in the Semantics phase.

The upsides of this patch are many. It is much easier to understand the
generated code and the compiler logic now. The ternary operator is much
simpler. The entire (very complex) "mergeTracker" function in the
verifier is no longer needed, and other simplifications to the verifier
were made as well.

### DIFF
--- a/compiler/assembler.cpp
+++ b/compiler/assembler.cpp
@@ -919,7 +919,8 @@ Assembler::Assemble(SmxByteBuffer* buffer)
     code->header().flags = CODEFLAG_DEBUG;
     code->header().main = 0;
     code->header().code = sizeof(sp_file_code_t);
-    code->header().features = SmxConsts::kCodeFeatureDirectArrays;
+    code->header().features = SmxConsts::kCodeFeatureDirectArrays |
+                              SmxConsts::kCodeFeatureHeapScopes;
     code->setBlob(cg_.code_ptr(), cg_.code_size());
 
     // Set up the data section. Note pre-SourceMod 1.7, the |memsize| was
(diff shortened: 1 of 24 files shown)


SlidyBat (commented):
Shouldn't the first `expr` also be added to the `CommaExpr`?

dvander:
Thanks, I'll fix this.

PR #698, opened 2021-11-05
Implement RTTI parsing and validation in the VM

peace-maker:
This loads and validates all available `rtti.*` sections in the binary.

The new `.dbg.*` sections are only validated if the line debugger was enabled using `ISourcePawnEnvironment::EnableDebugBreak()` which isn't called in SourceMod yet.

The legacy `.dbg.symbols` section isn't parsed (yet?).

It's the first step to expose a symbol access API in the VM to implement an useful plugin debugger (ref #539).

### DIFF
--- a/compiler/assembler.cpp
+++ b/compiler/assembler.cpp
@@ -486,6 +486,7 @@ RttiBuilder::add_enumstruct(Type* type)
         info.type_id = to_typeid(encoding);
         info.offset = field->addr();
         es_fields_->at(es.first_field + index) = info;
+        index++;
     }
 
     return es_index;
--- a/vm/rtti.h
+++ b/vm/rtti.h
@@ -1,0 +1,120 @@
+// vim: set sts=2 ts=8 sw=2 tw=99 et:
+//
+// Copyright (C) 2004-2021 AlliedModers LLC
+//
+// This file is part of SourcePawn. SourcePawn is licensed under the GNU
+// General Public License, version 3.0 (GPL). If a copy of the GPL was not
+// provided with this file, you can obtain it here:
+//   http://www.gnu.org/licenses/gpl.html
+//
+#ifndef _include_sourcepawn_rtti_h_
+#define _include_sourcepawn_rtti_h_
+
+#include <memory>
+#include <vector>
+
+#include <sp_vm_types.h>
+#include <amtl/am-hashmap.h>
+
+namespace sp {
+namespace debug {
+
+class Rtti;
+
+class RttiData {
+public:
+  RttiData();
+  RttiData(const uint8_t* blob, uint32_t size);
+
+  const Rtti* typeFromTypeId(uint32_t type_id) const;
+  const Rtti* functionTypeFromOffset(uint32_t offset) const;
+  const Rtti* typesetTypeFromOffset(uint32_t offset) const;
+
+  const uint8_t* blob() const;
+  size_t size() const;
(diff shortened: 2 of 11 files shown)


peace-maker:
😄 This caught the compiler emitting bad rtti data already.

compiler/assembler.cpp
```diff
@@ -485,7 +485,7 @@ RttiBuilder::add_enumstruct(Type* type)
         info.name = names_->add(field->nameAtom());
         info.type_id = to_typeid(encoding);
         info.offset = field->addr();
-        es_fields_->at(es.first_field + index) = info;
+        es_fields_->at(es.first_field + index++) = info;
```
dvander, line 488: Move the index++ onto a separate line for readability

vm/smx-v1-image.cpp
```diff
@@ -394,9 +399,154 @@ SmxV1Image::validateRtti()
...
+      return error("invalid enum name");
+  }
+  return true;
+}
+
+bool
+SmxV1Image::validateRttiEnumStructs()
+{
+  if (!rtti_enumstruct_fields_)
+		return error("rtti.enumstruct_fields section missing");
```
dvander, line 469: Looks like a hard tab leaked in here.

vm/rtti.h
```diff
@@ -0,0 +1,119 @@
...
+
+#include <sp_vm_types.h>
+#include <amtl/am-hashmap.h>
+
+namespace sp {
+
+  class Rtti;
+  class SmxV1Image;
+
+  class RttiData {
```
dvander, line 24: It doesn't look like this file is used anywhere in the tree yet? Does it need to be in this changeset?

peace-maker: It's used to validate the type information in all the other rtti sections, but the `Rtti` class isn't used yet indeed. I could strip out the decoding functionality and only leave the validation for now, if that seems better.

dvander: Ack, I missed where it's #included. It's fine for now, though the body should be dedented (namespaces don't increase the indent level).

It'd be better if it was clearly marked as something used for debugging purposes though. Slapping a "debug" namespace around it might be best. The reason is that, I haven't fully mapped out where RTTI is going. We don't currently use it for anything, and it's not clear it would exist in a source-only environment. And if it does, `Rtti` almost certainly won't work for actual runtime stuff.

I don't want to hold up your reviews trying to refactor something for a completely unknown use case, but I also don't want to have to refactor a bunch of dependent code later too.

So keeping it inside a namespace just for its usecase for now seems like the best path forward.

peace-maker: Understood. I don't want to pin the binary layout of the rtti/debug info, but that's a necessary evil if we want some state introspection for dynamic debugging (which I really want :D). Since you didn't use the info yet, I didn't push eagerly for these changes to be merged and it always sounded like you're about to use them in some way which could mean changing the layout again.

Can't talk about switching the architecture for a source-only runtime, since that's all on your agenda in the upcoming decades. We can always drop support for debugging again, if it's too much of a hassle.

The Rtti classes live in the `sp::debug` namespace now and I'll add all the other code regarding debugging support in there as well.

dvander:
Thanks! LGTM, I'm going to run this against the corpus before submitting.

dvander:
No new verification failures observed over 1.10-compiled binaries.

PR #710, opened 2021-11-15
Fix do-while flow compilation

PerfectLaugh:
This fixes incorrect behavior of do-while which causing infinite loop when encountering continue statement.

### DIFF
--- a/compiler/code-generator.cpp
+++ b/compiler/code-generator.cpp
@@ -1548,15 +1548,19 @@ CodeGenerator::EmitDoWhileStmt(DoWhileStmt* stmt)
     loop_cx.heap_scope_id = heap_scope_id();
     ke::SaveAndSet<LoopContext*> push_context(&loop_, &loop_cx);
 
-    __ bind(&loop_cx.continue_to);
-
     auto body = stmt->body();
     auto cond = stmt->cond();
     if (token == tDO) {
+        Label start;
+        __ bind(&start);
+
         EmitStmt(body);
-        if (!body->IsTerminal())
-            EmitTest(cond, true, &loop_cx.continue_to);
+
+        __ bind(&loop_cx.continue_to);
+        EmitTest(cond, true, &start);
     } else {
+        __ bind(&loop_cx.continue_to);
+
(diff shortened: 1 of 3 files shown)


dvander:
Thanks for the fix - can you include a test case?

PerfectLaugh:
Before:
image

After:
image

Fyren:
I think the test for `do` always needs to be emitted, regardless of `IsTerminal`.  If not, a `continue` in the loop body will just end the loop.

```
int x = 0; 
do { 
  x++;
  continue; 
} while (x < 3);
```

PerfectLaugh:
> I think the test for `do` always needs to be emitted, regardless of `IsTerminal`. If not, a `continue` in the loop body will just end the loop.
> 
> ```
> int x = 0; 
> do { 
>   x++;
>   continue; 
> } while (x < 3);
> ```

Thanks for your kindly correction. I didn't know about what IsTerminal means so I made the mistake.

tests/basic/while.sp
```diff
@@ -10,4 +10,31 @@ public main()
...
+        break;
+      }
+      else {
+        print("even\n");
+      }
+
+      continue;
+    }
+
+    print("old\n");
```
KyleSanderson, line 30: Should this say odd?

dvander: I'm holding off merging until I can run it against the corpus, which will probably be a day or two.

If PerfectLaugh wants to fix the typo before then, it'd be good to also split these two new cases into separate test files. Debugging tests is much easier when they're small, so we try not to make a single test function too big.

But I'm okay to merge this either way, typo fixed and tests split or not, I can fix that later.

dvander (approved):

tests/basic/while.sp
PerfectLaugh: Sorry my English is bad ;\

@dvander I migrated the do-while part into separate file named "do-while.sp"

dvander:
Thanks! I've manually merged this to master, with a few tiny fixes (split up the test further, and retained a codegen optimization where we don't have to emit on break/returns).

PR #717, opened 2021-11-19
Fix array generation and recycling

PerfectLaugh:
Previous HeapAlloc2dArray failed under HeapPop due to incorrect heap layout.

### DIFF
--- a/vm/plugin-context.cpp
+++ b/vm/plugin-context.cpp
@@ -1023,7 +1023,7 @@ bool
 PluginContext::usesHeapTracker() const
 {
   LegacyImage* image = runtime()->image();
-  return !!(image->DescribeCode().features & SmxConsts::kCodeFeatureHeapScopes);
+  return !(image->DescribeCode().features & SmxConsts::kCodeFeatureHeapScopes);
 }
 
 bool
--- a/vm/shell.cpp
+++ b/vm/shell.cpp
@@ -241,7 +241,10 @@ static cell_t Copy2dArrayToCallback(IPluginContext* cx, const cell_t* params)
   fn->PushCell(params[3]);
   fn->Execute(&ignore);
 
-  cx->HeapRelease(addr);
+  if ((err = cx->HeapPop(addr)) != SP_ERROR_NONE) {
+    return cx->ThrowNativeError("HeapPop error %d", err);
+  }
+
   return 0;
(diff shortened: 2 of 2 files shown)


vm/plugin-context.cpp
```diff
@@ -497,7 +497,6 @@ PluginContext::popTrackerAndSetHeap()
 int
 PluginContext::pushTracker(uint32_t amount)
 {
-  assert(usesHeapTracker());
```
dvander, line 500: The tracker must not be used in >= 1.11 binaries. Can you clarify what is being fixed here? I didn't see an explanation or test case.

vm/shell.cpp
```diff
@@ -233,15 +233,17 @@ static cell_t Copy2dArrayToCallback(IPluginContext* cx, const cell_t* params)
 
   cell_t addr;
   if (!cx->HeapAlloc2dArray(params[2], params[3], &addr, flat_array))
-    return 0;
+    return cx->ThrowNativeError("Could not allocate %d * %d 2D Array: %d", params[2], params[3], err);
```
dvander, line 236: Functions returning "err" have already called ReportError, and don't have an error code.

vm/plugin-context.cpp
PerfectLaugh: The problem is HeapPop doesn't deallocate the resource allocated by HeapAlloc2dArray

https://github.com/alliedmodders/sourcemod/blob/fix-menu-vote-handler/core/logic/smn_menus.cpp#L510

Returns error 5. SP_ERROR_INVALID_ADDRESS.

PerfectLaugh: image
This is the error from mapchooser_extended after running a while

vm/shell.cpp
PerfectLaugh: Sorry I forgot to remove the left messages

vm/plugin-context.cpp
PerfectLaugh: I didn't know the tracker usage.

dvander: Okay, thanks. This API does feel very weird to me. I'll add an RAII thing (eg "AutoHeapScope") to make this way easier to use. The current API is too difficult.

PR #730, opened 2021-11-29
Exact match check for #pragma newdecls values

maxijabase:


dvander:
This broke a very large number of plugins in our test corpus (which is currently not used in CI but I'm working on that). It turns out this is a common pattern:

    #pragma newdecls required;

And the semicolon at the end breaks this. If you want a go at another fix, test cases would be good. Note that the following is also allowed right now:

    #pragma semicolon 1;

So the implementation of that may serve as a guide.

PR #731, opened 2021-11-29
Fix incorrect continue behavior on always-exits for loop

PerfectLaugh:
DNANRH%B%QIE9)XZZO70$9
This commit fixes lost continue jump on above code.


PerfectLaugh:
Test is failed because github.com connection failure

compiler/code-generator.cpp
```diff
@@ -1619,6 +1619,7 @@ CodeGenerator::EmitForStmt(ForStmt* stmt)
     auto body = stmt->body();
     bool body_always_exits = body->flow_type() == Flow_Return ||
                              body->flow_type() == Flow_Break;
+    body_always_exits = body_always_exits && !(stmt->has_continue());
```
dvander, line 1622: This would read better as:

    bool body_always_exits = false;
    if (body->flow_type() == Flow_Return || body->flow_type() == Flow_Break) {
        if (!stmt->has_continue())
            body_always_exits = true;
    }

dvander:
Thanks for adding tests!

dvander:
Thanks for finding and fixing this - I'll pin a new SP to sourcemod in the next day or two.

PR #738, opened 2021-12-10
Fix top-level cxx polution

Headline:
This patch should fix some redefinitions we were seeing in unrelated areas upstream in SourceMod by setting these flags on the binary and not cxx, which should only be used directly by top-level scripts.


dvander (approved):
Doh. Thanks for finding this.

PR #739, opened 2021-12-11
Native parameters shouldn't shadow anything.

Fyren:


compiler/symbols.cpp
```diff
@@ -325,6 +325,10 @@ enum class NewNameStatus {
 static NewNameStatus
 GetNewNameStatus(SemaContext& sc, sp::Atom* name, int vclass)
 {
+    // Native parameters shouldn't shadow
+    if (sc.func_node() && sc.func_node()->is_native())
```
dvander, line 329: Is this supposed to be legal?

    native void Blah(int a, int a, int a);

dvander (approved):

PR #744, opened 2021-12-13
Fix dynamic array gen with old syntax.

Fyren:


asherkin:
@dvander does it make sense to merge this and open a new issue for the ASAN issue with the missing error messages?

dvander:
I'd rather not merge something that causes tests to fail. I'll take a look soon.

PR #756, opened 2022-01-22
Support ASAN option for Windows

Headline:
This was a lot less work than what we had to do for clang, we just have to check msvc version

### DIFF
--- a/AMBuildScript
+++ b/AMBuildScript
@@ -241,11 +241,15 @@ class Config(object):
         ]
 
         if self.enable_asan:
-            raise Exception('--enable-asan only supported when using Clang')
+            self.configure_asan(cxx)
 
     def configure_asan(self, cxx):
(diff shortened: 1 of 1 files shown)


AMBuildScript
```diff
@@ -241,11 +241,17 @@ class Config(object):
...
+                raise Exception('--enable-asan requires Visual Studio version >=16.9')
+            self.configure_asan(cxx)
 
     def configure_asan(self, cxx):
-        if cxx.target.platform != 'linux':
-            raise Exception('--enable-asan only supported on Linux')
+        if cxx.target.platform == 'windows':
+            cxx.cflags += ['/fsanitize=address']
+            return
+        elif cxx.target.platform != 'linux':
```
dvander, line 252: nit: no need for "elif" here, "if" will suffice

dvander (approved):

PR #767, opened 2022-02-10
Add a few plugins to corpus manifest

JoinedSenses:
Add a few public plugins that make use of some newer language features or might contain unusual formatting

### DIFF
--- a/tools/corpus/manifest.yaml
+++ b/tools/corpus/manifest.yaml
@@ -5,6 +5,9 @@ git:
         - https://github.com/dysphie/sourcemod-indev incap-wip.sp
         - https://github.com/dysphie/nmrih-qol-ex scripting/qol.sp
         - https://github.com/Bara/fuckZones addons/sourcemod/scripting/fuckZones.sp
+        - https://github.com/JoinedSenses/TF2-Stasis scripting/stasis-dev.sp
+        - https://github.com/JoinedSenses/SM-ShowTriggers-Redux scripting/showtriggers.sp
+        - https://github.com/JoinedSenses/SM-ConnectionData scripting/connectiondata.sp
 
 includes:
         - https://github.com/alliedmodders/sourcemod plugins/include


JoinedSenses:
Im uncertain of the behavior of connectiondata here due to it using geoip2, which has since been added to sourcemod

dvander:
This entry fails to compile:

    github.com/JoinedSenses/TF2-Stasis/scripting/stasis-dev.sp(200) : error 413: returned array does not have the same dimension count as return type

JoinedSenses:
> This entry fails to compile:
> 
> ```
> github.com/JoinedSenses/TF2-Stasis/scripting/stasis-dev.sp(200) : error 413: returned array does not have the same dimension count as return type
> ```

This error does not make sense and compiles fine in 1.10. The func makes an enum struct and then returns it. Is it a problem with declaring the ret as `Player[]`, particularly []? Otherwise this method doesnt compile.

dvander:
The <= 1.10 compilers would ignore the "[]" on the return type and then auto-infer whether it was an array or not. 1.11 requires that the return type and actual returned values are compatible types.

dvander:
It's an easy fix, from looking at the code - just change the return type to "Player". I believe this was noted in the 1.11 breaking changes doc, since it did affect a few plugins on the forums.

JoinedSenses:
Sure, thanks. I'll update at some point, maybe whenever 1.11 is marked as stable. In 1.10, removing the brackets spews a ton of errors for the above code.

> plugin.sp(10) : error 001: expected token: ";", but found "("
plugin.sp(12) : error 021: symbol already defined: "p"
plugin.sp(13) : error 010: invalid function or declaration
plugin.sp(18) : warning 219: local variable "p" shadows a variable at a preceding level
plugin.sp(19) : error 012: invalid function call, not a valid address
plugin.sp(19) : warning 215: expression has no effect
plugin.sp(19) : error 001: expected token: ";", but found ")"
plugin.sp(19) : error 029: invalid expression, assumed zero
plugin.sp(19) : fatal error 190: too many error messages on one line

dvander:
lol - if I had to guess, the 1.10 has a broken newdecl parser for return types, and it falls back to trying a variable declaration.

Issue #786, opened 2022-04-09
Tag-mismatch shows incorrect line


Vinillia:
<details>
  <summary>SourceMod version</summary>

```
SourceMod Version Information:
SourceMod Version: 1.11.0.6863
SourcePawn Engine: 1.11.0.6863, jit-x86 (build 1.11.0.6863)
SourcePawn API: v1 = 5, v2 = 16
Compiled on: Mar 10 2022 01:11:20
Built from: https://github.com/alliedmodders/sourcemod/commit/01203a5a
Build ID: 6863:01203a5a
http://www.sourcemod.net/
```

</details>

<details>
  <summary>Code</summary>

```
#pragma newdecls required
#pragma semicolon 1

#include <sourcemod>

int i;
bool b;
Handle h;

public void OnPluginStart()
{
	if ( i > b ){}
	if ( i > h ){}
	if ( b > h ){}
	
	i > b;
	i > h;
	b > h;
}

void nothing()
{
	
}
```

</details>

If you try to compile code above you will get bunch of warnings and as you can see all warning matches their lines except for "tag mismatch". For some unknown for me reason compiler thinks that all tag mismatches is on 24 line which is definitely some mistake.

```
test3.sp(16) : warning 215: expression has no effect
test3.sp(17) : warning 215: expression has no effect
test3.sp(18) : warning 215: expression has no effect
test3.sp(21) : warning 203: symbol is never used: "nothing"
test3.sp(24) : warning 213: tag mismatch (expected "int", got "bool")
test3.sp(24) : warning 213: tag mismatch (expected "int", got "Handle")
test3.sp(24) : warning 213: tag mismatch (expected "bool", got "Handle")
test3.sp(24) : warning 213: tag mismatch (expected "int", got "bool")
test3.sp(24) : warning 213: tag mismatch (expected "int", got "Handle")
test3.sp(24) : warning 213: tag mismatch (expected "bool", got "Handle")
```

Fyren:
At least the specific bug the example SP code hits is fixed, which involved tag mismatches with relational operators, but the fix should be more general.  If anyone finds more mis-reported line numbers, please file a new bug with sample code.  (Make sure you're using a version of spcomp with the fix since SM doesn't automatically build against the latest SP commit.)

PR #794, opened 2022-04-30
Fix wrong line reporting for errors (issue #786)

Fyren:
Bug: issue #786
Test: regressions/fail-error-reporting-pos.sp

Also modifies `runtests.py` to allow passing `type` through from the
test file into the manifest.


dvander:
Looks good, thanks!

PR #795, opened 2022-04-30
Fix peek_same_line

Fyren:
Fix peek_same_line when the next token is in a different file, but happens to have the same line number.


dvander:
LGTM

PR #821, opened 2022-07-04
Fix string array dimensions in rtti

peace-maker:
The `* 4` of the last dimension appears to be left over after some prior refactoring and is not needed anymore.


dvander:
lgtm feel free to cp directly to 1.11-dev if you need it.

Issue #823, opened 2022-07-08
[Feature Request] static_assert should work in local scopes


sirdigbot:
Currently, `static_assert()` only works in the global scope. To me this is a pointless limitation. Ideally it should be usable in functions to let you know exactly *where and why* an assertion is failing.

If a function doesn't compile it's because it's not written correctly. So naturally if an assert *inside* of a function fails it implies the function itself is not written correctly--it's making incorrect assumptions.

An example that I've run into a few times
```sourcepawn
enum ThingType
{
    ThingType_A = 0,
    ThingType_B,
    ThingType_C, // We've just added this new type!
    ThingType_Total
};

// Meanwhile, in one of many similar places in the codebase
void DoThingAccordingToType(ThingType t)
{
    // error 029: invalid expression, assumed zero
    // static_assert(ThingType_Total == 2, "Forgot to implement new ThingType");

    if (t == ThingType_A)
        GiveDiploma();
    else if (t == ThingType_B)
        SellKidneysToOrphanage();
    else
        FailSilently();
}
```
It would be nice to be able to use static_assert inside of this function. It's not a bullet-proof fix by any means but it's more sensible than a runtime check of the same condition.

Or at the very least fix the error message ¯\\\_(ツ)_/¯

Currently the closest workaround is this template-looking function definition, which is gross.
```sourcepawn
static_assert(ThingType_Total == 2, "Forgot to implement new ThingType");
void DoThingAccordingToType(ThingType t)
{
    // ...
}
```

Fyren:
Added to master.

PR #827, opened 2022-07-08
Allow static_assert as a statement.

Fyren:
Bug: issue #823
Tests: compile-only/fail-static-assert-message.sp
       compile-only/fail-static-assert-no-message.sp
       compile-only/ok-static-assert-enum-value.sp


dvander:
lgtm, thanks!

Issue #841, opened 2022-08-09
Deleting array of handles without index doesn't emit compile error


Vinillia:
This code compiles just fine and causes runtime error. Tested on 1.12.0.6910 and 1.10.0.6545 compiler versions
```
Handle timers[200];

public void OnPluginStart()
{
    delete timers;
}
```
```
Exception reported: Handle 8e8 is invalid (error 4)
Blaming: issue.smx
Call stack trace:
  [0] CloseHandle
  [1] Line 5, c:\Users\statk\Desktop\drop_down_fix\issue.sp::OnPluginStart
```

PR #842, opened 2022-08-10
Fix array handling for delete statements (#841)

Fyren:
I think the intent with the `dim.array.level` check was to error if the expression wasn't the last dimension in the array, but `level` holds how many levels are below.  So for something like `int arr[2][2][2]`, both `arr[0][0]` (which is like `int[2]`) and `arr[0][0][0]` (which is like `int`) have a `level` of 0.

Removing `iARRAYCHAR` and `iARRAYCELL` just let it go on and check if the actual type can be deleted for such expressions.  I don't know what kind of expression would have led to a null `sym`, though.


dvander (approved):
Thanks!

Issue #849, opened 2022-09-29
Internal compilation error when declaring a big array


misdocumeno:
when compiling this
```cpp
public void OnPluginStart() {
    int a[536870908];
}
```
the compiler output is this 
```
SourcePawn Compiler 1.11.0.6911
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2021 AlliedModders LLC

test.sp(2) : warning 203: symbol is never used: "a"
Binary validation failed: Method OnPluginStart failed verification: Instruction contained invalid parameter

Internal compilation error detected. Please file a bug:
https://github.com/alliedmodders/sourcepawn/issues/new
```

ZekyTheWolf:
Interesting, windows/ubuntu 22 worked... 

But why would you need that big size?

```c
#include <sourcemod>

#pragma newdecls required
#pragma semicolon 1

public void OnPluginStart() {
    char a[536870908];

}
```
Console (Same output for Windows/Ubuntu)

```
SourcePawn Compiler 1.11.0.6911
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2021 AlliedModders LLC

e:\SourceMod\test.sp(7) : warning 203: symbol is never used: "a"
Code size:         3560 bytes
Data size:         2280 bytes
Stack/heap size:   536887356 bytes
Total requirements:536893196 bytes

1 Warning.
```

Poggicek:
Note he said spcomp64, 32 bit version doesn't reproduce this

ZekyTheWolf:
> Note he said spcomp64, 32 bit version doesn't reproduce this

Both, spcomp & spcomp64 compiling it without any problem.

misdocumeno:
just messing around with the compiler, i dont really need that size, but since ints a bug, i figured it might be worth creating an issue

it's spcomp64 on arch linux, tho i just realized that with the 32 bit version, a different error happens

```
SourcePawn Compiler 1.11.0.6911
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2021 AlliedModders LLC

test.sp(2) : warning 203: symbol is never used: "a"
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```

Mooshua:
This is not a bug. You are overflowing the allocation size parameter in the allocation opcodes.

[...]

Issue #850, opened 2022-10-01
#pragma bugs detection of inconsistent indentation


dragokas:
The following code causing warning:
```
void SampleTranslation(int client = LANG_SERVER)
{
	#pragma unused client
	PrintToChatAll("Hello World");
}
```
```
warning 217: inconsistent indentation (did you mix tabs and spaces?)
```
despite only tabs are used.

dvander:
I can't reproduce this on 1.11 - what compiler did you test on?

dvander:
Thanks, this has been fixed on master.

dragokas:
dvander,
```
SourcePawn Compiler 1.11.0.6905
Copyright (c) 1997-2006 ITB CompuPhase
Copyright (c) 2004-2021 AlliedModders LLC

C:\Users\Alex\Desktop\1.sp(1) : warning 203: symbol is never used: "SampleTranslation"
C:\Users\Alex\Desktop\1.sp(4) : warning 217: inconsistent indentation (did you mix tabs and spaces?)
Code size:         3488 bytes
Data size:         2280 bytes
Stack/heap size:      16452 bytes
Total requirements:   22220 bytes

2 Warnings.
```
I used tabs.

dvander:
Can't repro on Windows either.

PR #854, opened 2022-10-12
Fix codegen error on huge array allocation.

dvander:
Bug: issue #850
Test: test suite


peace-maker:
This fixes #849 not 850

PR #869, opened 2023-01-22
Remove usage of sprintf on newer apple-clang

peace-maker:
note: 'sprintf' has been explicitly marked deprecated here __deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")

SourceMod Scripting CI failed on mac due to this.
https://github.com/alliedmodders/sourcemod/actions/runs/3979033169/jobs/6821186383

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -1918,15 +1918,15 @@ bool Lexer::need(int token) {
 void Lexer::NeedTokenError(int token, int got) {
     char s1[20], s2[20];
     if (token < 256)
-        sprintf(s1, "%c", (char)token); /* single character token */
+        SafeSprintf(s1, sizeof(s1), "%c", (char)token); /* single character token */
     else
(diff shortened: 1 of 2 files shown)


psychonic (approved):

Wend4r:
Maybe this is the best solution for OS X
https://github.com/Wend4r/sourcemod/commit/fe0363eeea125e8697dd54336883c1bd6cd3ca4c

I think If you translate the project completely to AMTL, then it must be purposeful, removes all usage of C Library (or some of its replacement components)

dvander:
Wend4r: just pleasing the compiler is fine for these kinds of fixes. There is no actual bug, and the code is working fine either way.

Wend4r:
I think for OS X it is preferable to use the C Library because according to Apple Rosetta (2) technology, the code is being accelerating at such moments (by emulating x86_64 to ARM64/PowerPC) (possibly a request to AMTL).

PR #887, opened 2023-07-13
Should see predefined symbols as macros

PerfectLaugh:
The use case of `spcomp test.sp SYM=` should be seen as macros instead of constants in our use cases. Thus this PR.

```sourcepawn
#if defined SYM
	#define A 11
#endif

void func() {
#if A
	PrintToServer("A");
#endif
}
```


dvander:
Previously these were not macros, they were global constants. However the preprocessor used to be able to see global constants which is no longer allowed. Defining constants on the command line seems dumb, so I'm in favor of this change.

However, it does need to include a new test suite entry. You can add a new metadata key here:

https://github.com/alliedmodders/sourcepawn/blob/master/tests/runtests.py#L254

This should enable you to do something like this at the top of your test:

    // define: X=Y

Then here, in run_compiler:

https://github.com/alliedmodders/sourcepawn/blob/master/tests/runtests.py#L473

You can add the arguments from the test metadata.

PerfectLaugh:
> Previously these were not macros, they were global constants. However the preprocessor used to be able to see global constants which is no longer allowed. Defining constants on the command line seems dumb, so I'm in favor of this change.
> 
> However, it does need to include a new test suite entry. You can add a new metadata key here:
> 
> https://github.com/alliedmodders/sourcepawn/blob/master/tests/runtests.py#L254
> 
> This should enable you to do something like this at the top of your test:
> 
> ```
> // define: X=Y
> ```
> 
> Then here, in run_compiler:
> 
> https://github.com/alliedmodders/sourcepawn/blob/master/tests/runtests.py#L473
> 
> You can add the arguments from the test metadata.

Added test case per requested

https://github.com/alliedmodders/sourcepawn/pull/887/files#diff-dfd4f392838cd2972651a8d8831ebf0faa471ee85dc2220efa86ed7a946f3012R1

dvander:
Looks good, but the diff in runtests.py is huge - was there a line endings change by accident?

PerfectLaugh:
> Looks good, but the diff in runtests.py is huge - was there a line endings change by accident?

Didn't make my autocrlf work. Now the diff should look normal.

dvander:
Thanks!

PR #900, opened 2023-09-30
Add beautiful error messages.

dvander:
This is easy now that we have proper location encoding.

The current formatting is debatable, but it's a start.
 Example:

    test.sp(3) : error 017: undefined symbol "ham"
         3 |     int hello = ham;
    -------------------------^

    test.sp(4) : error 017: undefined symbol "hhello"
         4 |     return hhello
    --------------------^

### DIFF
--- a/compiler/errors.cpp
+++ b/compiler/errors.cpp
@@ -169,6 +169,7 @@ MessageBuilder::~MessageBuilder()
     auto& cc = CompileContext::get();
 
     ErrorReport report;
+    report.loc = where_;
     report.number = number_;
     if (where_.valid())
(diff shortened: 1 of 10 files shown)


assyrianic:
how does this handle macros?

dvander:
It'll point at the expansion location in the source. We can actually do much better, since we retain the expanded macro text, and column information within it. However, I don't think I'm going to actually make use of this.

Macros are really complicated, entirely unnecessary, and very expensive in both memory use and parsing time. To make things worse, the vast majority of uses are single token replacement, which doesn't benefit from any additional diagnostics. So basically they're bad in every single way and we'll all be better off by pretending they don't exist.

PR #903, opened 2023-10-02
Remove the indentation warning. 

dvander:
In the early days of AMX Mod X/SourceMod there weren't really good
source editors. Many people used Notepad for their scripting. Mixing
tabs/spaces and getting indents wrong was very common.

These days there are a few factors that make this less critical:
  - There are many high quality editors available, and there is much
    more awareness around them.
 - The industry has basically settled on spaces for indentation.
 - Modern programming languages tend to come with auto-formatting tools now as
   well, which is probably something SourcePawn should do.

Either way, the inline linter is a huge, huge hack and has no business
being right there in the lexer. It makes token caching very complex.
It's slow. And it requires knowing a predefined value for tabsizes,
which makes no sense.

So awaaaay it goes.

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -782,9 +782,6 @@ bool Lexer::FindNextToken() {
         auto work_line = line_start();
         bool is_line_start = (work_line == work_start) && !IsInMacro();
 
-        if (is_line_start)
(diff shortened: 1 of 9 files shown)


psychonic:
Purge the unclean.

PR #912, opened 2023-10-09
Remove nullfunc_t and make INVALID_FUNCTION the same as null.

dvander:
INVALID_FUNCTION was proving to be a major pain; setting it to -1 might be one of the dumbest things I've done. It had a few problems all boiling down to "there is no support in the compiler for non-zero false values". Eg, `f1 == f2` is false below:

        Function f1;
        Function f2 = INVALID_FUNCTION;

We could have gone through and fixed all the edge cases like this throughout the compiler, but it seems counter-productive.

The other oddity is that because `null` is 0, `INVALID_FUNCTION` needed a separate unit type, which we called `nullfunc_t`.

Rather than keep propping up all this weirdness, this patch normalizes `INVALID_FUNCTION` by making it equivalent to `null`. Its value is 0, and its type is `null_t`.

This will break compatibility with cross-plugin interactions, so plugins sharing a scripted API will need to be recompiled.

It will also break C++ code relying on a -1 value. To help assist with this breakage, three functions have been added to `IPluginContext`:

 - `GetNullFunctionValue()`
 - `IsNullFunctionId()`
 - `GetFunctionByIdOrNull()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,3 +185,14 @@ Also as of this release, both the compiler and VM can run on non-x86 platforms
 ### SourcePawn 1.12
 
 SourcePawn 1.12 is currently in development.
+
+This release contains a number of bug fixes and internal improvements. It also greatly improves
+error message readability in the style of modern compilers.
+
+This release contains a number of language changes.
+ - Function bodies must now be braced. There is no compatibility mode for this change.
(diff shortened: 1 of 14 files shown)


Sikarii:
I take this fixes #469 so the issue can most likely be closed.

PR #916, opened 2023-10-22
Refactor how builtin AST nodes are created.

dvander:
Having custom C++ to create built-in symbols introduced a lot of weird complexity around scopes and source locations. It was also a cludge, because it meant the compiler could rely on idiosyncracies that are not exposed to the language itself (eg, INVALID_FUNCTION was a type completely inaccessible to scripts).

It also posed another problem. For historical reasons the compiler still separates the concept of "AST Node" and "Symbol", and trying to merge them is proving very difficult to do in small steps. Builtin symbols make it even more difficult because they didn't have corresponding AST nodes.

The new mechanism for builtins is to embed SourcePawn-language scripts into C++ code as strings, and then inject those strings into the parser. This is very easy to do thanks to all of our recent refactorings. And, again thanks to recent refactorings, we even get correct file/line information if these builtins happen to have an error.

There are some user-visible changes as a result of this.

1. Command-line macros (eg spcomp.exe X=Y) will now create an in-memory script, like:

```
     #define X Y
```

  This will improve diagnostics a bit.

2. Built-in defines and constants are now created via a separate in-memory script, that looks like this:

```
     #define __BINARY_PATH__ "test.smx"
     #define __BINARY_NAME__ "test.smx"
     #define __DATE__ "10/21/2023"
     #define __TIME__ "23:19:59"
     const int EOS = 0;
     const int cellmax = 2147483647;
     const int cellmin = -2147483648;
```

3. The "using" keyword has been removed and is no longer implemented. The __nullable__ and destructor syntax for handles has been re-introduced, and SourceMod will have to bring back the Handle methodmap. Because the native alias syntax no longer exists, CloseHandle will not suffice, and SourceMod will also have to add a "Handle.~Handle" native mapping.


dvander:
This is a catch-22 to merge because it needs an SM update to pass the corpus. So for now, I'll run the corpus locally.

PR #938, opened 2023-12-11
Unused arguments in stock functions are now triggering warnings

### LINKED ISSUE #784: "Stock" hides unused variable warning
I'm not sure if this is a bug but it seems like one.
The `stock` storage class specifier (are they called that in SP?) suppresses unused variable warnings on functions that *are* being used.
`public` does too but I think that's intentional.
```sourcepawn
public void OnPluginStart()
{
    int x;
    int y;
    char z[32];

    Func1(x);
    Func2(x);
    Func3(x, y, z);
    Func4(x, y, z);
    
    FuncOther(x);
}


void Func1(int a) // warning 203: symbol is never used: "c"
{
    PrintToServer("side-effect");
    return;
}

stock void Func2(int a) // No warning
{
    PrintToServer("side-effect");
    return;
}


void Func3(int &a, int &b, char c[32]) // warning 203: symbol is never used: "c"
{
    PrintToServer("side-effect");
    return;
}

stock void Func4(int &a, int &b, char c[32]) // No warning
{
    PrintToServer("side-effect");
    return;
}


public void FuncOther(int a) // No warning (intentional?)
{
    PrintToServer("side-effect");
    return;
}
```

Side note, there's also no other way to make an array parameter optional (as in the function doesn't need to use the parameter, not that you don't need to pass the parameter), since unlike cell types you can't make it a reference to dismiss the warning.

elymination:
These changes fix #784

`tests/compile-only/warn-unused-arg.sp` adds coverage for unused arguments in `public`, `static`, `stock` and regular functions

Note that `tests/compile-only/ok-argument-shadows-function.sp` had to be modified (`#pragma unused` added) to take into account that arguments should now be used in stock functions.

Feel free to tell me if I missed anything!

### DIFF
--- a/compiler/name-resolution.cpp
+++ b/compiler/name-resolution.cpp
@@ -892,7 +892,7 @@ FunctionDecl::BindArgs(SemaContext& sc)
 
         if (var->type_info().ident == iREFERENCE)
             var->set_is_read();
-        if (is_callback_ || is_stock_ || is_public_)
+        if (is_callback_ || is_public_)
             var->set_is_read();
 
         /* arguments of a public function may not have a default value */
--- a/compiler/semantics.cpp
+++ b/compiler/semantics.cpp
@@ -2494,9 +2494,9 @@ bool Semantics::TestSymbol(Decl* sym, bool testconst) {
         default: {
             auto var = sym->as<VarDeclBase>();
             /* a variable */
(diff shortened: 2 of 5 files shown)


dvander:
Thanks for the change, and for adding tests!

PR #945, opened 2024-02-08
Make enum structs into proper types.

dvander:
When we first added enum structs to the compiler, it was pre-AST, and
the only way we could get it done was by tossing in tons and tons of
hacks to desugar enum structs into arrays. It required introducing the
concept of "semantic tags" that represented the intended type of an
array, since it would get desugared into cells, and very often we need
to de-desugar (sugar?) back into enum structs.

These hacks are particularly gross because they don't add any real value
to the compiler other than pain.

We've delayed fixing this for a very long time because, to date, Pawn
has been all about cells, and not at all about types. But with the AST,
and with `struct symbol` removed, and numerical tags removed, we can
finally cast off these shackles.

This is the first, official, truly non-cell type in Pawn. It's still got
plenty of hacks, but we've exposed all the nooks and crannies where
"variable size" equals "one cell", and that will prove very useful going
forward. This is also a starting point to eliminating iREFARRAY and
iARRAY which are pretty much complete nonsense.

This also means some changes are in order. A lot more semantic checks
need to make sure enum structs don't get bypassed. Calling matchtag()
and ignoring the result is no longer okay. We need additional checks to
find things matchtag() won't handle.

Anything that dealt with the size of an expression (eg, variable
declarations, sizeof, assignment) needs to check the underlying type
first, because it might be bigger than one cell.

Any kind of array code needed to be audited to see how it previously
interacted with enum structs.

Array and struct initialization continues to be handled by a common code
path since it's so similar.

Nested enum structs and enum structs with arrays had to be audited for
correctness.

sizeof() now takes a "hier1" expression, meaning a lot more stuff can
be tossed at it, and we can let the rest of the semantic pass calculate
sizes for us automatically.

There will be problems with this, and it'll take some time to root all
the corner cases out. But it's the right direction to go.


NotnHeavy:
This is some interesting stuff - will definitely experiment with this!

Funnily enough, I was working on a fork of SourcePawn recently as well to experiment with a potential `long` magic tag - whether it is within your interests or not, I can let you know how it goes. The code probably isn't perfect considering I haven't fully understood the SourcePawn repo (and things like codegen are more new to me) but it's still getting somewhere.

dvander:
If `int64` wasn't, well, 64-bits, I probably wouldn't be so sour on the magic tags route. There's an extra layer of hacks needed I'm not keen on.

Magic tags incur a huge amount of pain in the compiler. For something like "any" it's kind of tolerable because we could maybe imagine a type system where it can be consistently reasoned about, and right now that translates to a bunch of gross hacks in matchtag() and semantics.cpp. Sure, we could probably do that for other primitive types too. The problem is the VM which can't represent values > 32 bits.

That's fine for reference types, like enum structs, because internally the VM represents all addresses as 32-bit offsets relative from a base address. But for something like `int64` we'd need to make it a reference type in the VM but a value type in the language itself. That's another layer of hacks all over semantics.cpp and code-generator.cpp. I'm not at the point where I want to go down that road yet. I could see getting there, but *probably* not without eliminating `IdentifierKind`.

dvander:
This is looking better now. Still have some FIXMEs to address.

Issue #943, opened 2024-01-31
Preprocessor, "to string" conversion


Classes123:
I don't know if it's a bug or it was designed that way, but the following code gives the "wrong" result:
```cpp
#define TEST   15

#define _TO_STR(%0) #%0
#define TO_STR(%0)  _TO_STR(%0)

public void OnPluginStart()
{
    PrintToServer(TO_STR(TEST));
}
```
Result:
```
TEST
```
---
Same code in C:
```c
#define TEST   15

#define _TO_STR(x) #x
#define TO_STR(x)  _TO_STR(x)

int main()
{
    printf(TO_STR(TEST));
    return 0;
}
```
Result:
```
15
```

PR #947, opened 2024-02-14
Add macro to string conversion

Classes123:
The change allows you to convert a macro (which has no args) to string.

Related to alliedmodders/sourcepawn#943

---
#### Where can this be used?
In code/libraries, you can often see duplication of constants in numeric and string values (versions, parameters, etc). This creates inconvenience that one value needs to be edited in several places.
In general, if you want to quickly convert some predefined value to a string, then this can be done at the compilation stage.

---
#### A few notes:
I tried to implement "in-depth" macro parsing, but often I had to rewrite an existing function due to the inability to use custom streams. Perhaps such functionality is not needed in sp.
I also noticed that something similar had already been done in the experimental part, but for some reason it was abandoned.

imho, this is the simplest solution

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@
 *.suo
 *.user
 
+# IDEA
+.idea/
+
 # Build directories
 CrazyDebug*/
 Debug*/
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -2571,8 +2571,17 @@ std::string Lexer::PerformMacroSubstitution(MacroEntry* macro,
             out.push_back(substitute[pos + 1]);
             continue;
         }
-        if (stringize)
-            out += '"' + iter->second + '"';
+        if (stringize) {
+            out += '"';
+
+            auto arg_macro = FindMacro(cc_.atom(iter->second));
+            if (arg_macro != nullptr && !arg_macro->args)
+                out += arg_macro->substitute->str();
+            else
+                out += iter->second;
+
+            out += '"';
+        }
         else
             out += iter->second;
     }


dvander:
I'm very reluctant to take any changes to the macro parser. They should not be used. Need a strong justification as to why helper functions/normal variables don't work (with examples).Also need tests.On Feb 14, 2024 2:31 PM, Jolly Roger ***@***.***> wrote:The change allows you to convert a macro (which has no args) to string.
Related to #943

Where can this be used?
In code/libraries, you can often see duplication of constants in numeric and string values (versions, parameters, etc). This creates inconvenience that one value needs to be edited in several places.
In general, if you want to quickly convert some predefined value to a string, then this can be done at the compilation stage.

A few notes:
I tried to implement "in-depth" macro parsing, but often I had to rewrite an existing function due to the inability to use custom streams. Perhaps such functionality is not needed in sp.
I also noticed that something similar had already been done in the experimental part, but for some reason it was abandoned.
imho, this is the simplest solution



You can view, comment on, or merge this pull request online at:
  https://github.com/alliedmodders/sourcepawn/pull/947

Commit Summary
35c4576  Add macro to string conversion

File Changes  (2 files)

    M
    .gitignore
    (3)
  
    M
    compiler/lexer.cpp
    (13)
  

Patch Links:
https://github.com/alliedmodders/sourcepawn/pull/947.patchhttps://github.com/alliedmodders/sourcepawn/pull/947.diff

Classes123:
> I'm very reluctant to take any changes to the macro parser. They should not be used. Need a strong justification as to why helper functions/normal variables don't work (with examples). Also need tests.

okay,

1st example:
if I want to share the build of my plugin:
```cpp
// include/myplugin/info.inc
#define TO_STR(%0)  #%0

#define VER_A       1
#define VER_B       2

#define VER_STR     TO_STR(VER_A) ... "." ... TO_STR(VER_B) //now I don't have to duplicate the parameters above

// myplugin.sp
#include <myplugin/info>

public Plugin myinfo = {
    version = VER_STR
};

//...
```
of course I can do this by adding parameters to the compiler, having previously defined them in the environment, but it will be much easier this way.

---
2nd example:
Let's imagine that I have such a piece of code
```cpp
int g_state_a = 1;
int g_state_b = 2;
int g_state_c = 3;

//...
#define VAR_MAIN_STATE   g_state_a
//...

#define TO_STR(%0)  #%0
#define assert_main_state(%0) if (VAR_MAIN_STATE %0) SetFailState("State validation failed. (" ... TO_STR(VAR_MAIN_STATE) ...  " " ... #%0 ... ")")

public void OnPluginStart()
{
    //...
    assert_main_state(!= 5);
}
```
as you can see, with this design, I'm trying to print the value of the macro. In the current version of the compiler, I don't think this is possible.

The output will be `State validation failed. (g_state_a != 5)` instead of `State validation failed. (VAR_MAIN_STATE != 5)`

---
And yes, this feature is not very important, but it solves some problems.

PR #955, opened 2024-02-23
Remove iARRAY.

dvander:
This removes iARRAY as a storage class and makes arrays an actual built-in type. While this may seem boring, and has very little practical effect right now, it is the biggest change to the compiler we've done to date (including the refactoring to move to an AST). It fundamentally changes type checking, begins removing matchtag(), and begins removing Old Syntax.

Because arrays now fully participate in the type system, our type checks become much more comprehensive, and we can start building out unified type checking logic rather than hard-coding one-off checks all over the codebase. It's more work, but it increases the rigor of the type system, and it allows us to represent things we couldn't before.

For example, we can now have an iEXPRESSION that returns an array. Things like this are now possible:

    char c = "hello"[1];

This was not possible before, because expressions didn't have a way to attach array information other than via having a backing variable. That also means we can eliminate hacks like creating temporary Decls to represent intermediate values (like we used to in FieldAccessExpr).

As a consequence of now having a proper type checker, it is time to start removing "old" syntax - the pre-transitional, tag-based syntax. We will slowly begin removing matchtag() and turning type warnings into errors.

Type-related error messages have also been improved:

    error 047: array sizes do not match, or destination array is too small

Becomes:

    error 450: no viable conversion from "float[4]" to "float[3]"

In this change we are also removing support for tagged functions that return arrays. The following:

    String:blah() {
        char x[10] = "hello";
        return x;
    }

Is no longer valid. The return type must use typed syntax:

    char[] blah() {
        char x[10] = "hello";
        return x;
    }

Similarly, forgetting the brackets is also an error.


dvander:
This is getting close - tests pass, and the corpus is almost totally addressed. I'm hoping to merge it by EOW. A few additional notes are in order:

(1) What's the point? int64 support. Adding non-cell types will require correctly inserting coercion nodes in the AST. Coercion depends on type checking, which is what this change begins to add. 

(2) What are the next steps? Remove matchtag() and use the type checker everywhere. Then, introduce a coercion system, which will require either a pass that rebuilds the AST, or collapsing the existing passes into a single pipeline. Either way it's a big change that will take a while. Once that's in place, adding new types will be very simple.

(3) What about backward compatibility? We have to start breaking it, unfortunately. The type checker is uncovering straight up bugs in many plugins. It's not possible to keep every single quirk working and while adding new features, especially when the old quirks are so deeply broken.

(4) Will old decls go away? I'm 50/50 on this. A majority of plugins use old decls, and it mostly impacts the parser and not anything else (except for edge cases around arrays). My gut feeling is that it's better to pick away at the old syntax slowly rather than wholesale delete it. So, as the type checker improves, I'll be more aggressive about cutting old plugins loose.

Issue #976, opened 2024-06-06
[Compiler problem] Warning 203: symbol is never used


A1mDev:
A variable with the stock modifier now complains that the symbol is not used anywhere.
`Warning 203: symbol is never used "L4D2ZombieClassname"`

```
stock const char L4D2ZombieClassname[9][] =
{
	"smoker",
	"boomer",
	"hunter",
	"spitter",
	"jockey",
	"charger",
	"witch",
	"tank",
	"error_bad_L4D2ZombieClassType"
};

```

dvander:
Thanks, this is now fixed on master.

Issue #972, opened 2024-06-05
Static functions called by stock functions now report "symbol is never used" warnings


Mikusch:
Using the SourcePawn Compiler 1.12.0.7140, static functions will now report "symbol is never used" warnings if the stock they are referenced in is not used in the plugin. I'm not sure if this is the intended behavior now, but it doesn't make much sense in my opinion.

```sourcepawn
static void SomeHelperFunction()
{
	PrintToServer("I'm helping!");
}

stock void HelpMe()
{
	SomeHelperFunction();
}
```
Output:
```
spcomptest.sp(1) : warning 203: symbol is never used: "SomeHelperFunction"
     1 | static void SomeHelperFunction()
---------------------^
```

dvander:
In an include, this should not be an error.

Mikusch:
> In an include, this should not be an error.

The warning does show up in includes.

PR #979, opened 2024-06-09
Fix for #976 to Ignore Unused Stock Variables

c0rp3n:
# Solution

This change looks to ignore stock variables that are not from the main source file.

# Testing

Added two new compile tests for stock variables and arrays.


dvander:
This error is intentional. The warning should only elide on includes.

A1mDev:
#976

c0rp3n:
Updated to elide the warning on includes.
I could also look to have the same behaviour for #972 but they would likely want to share code.

compiler/semantics.cpp
```diff
@@ -2307,6 +2307,18 @@ bool Semantics::CheckExprStmt(ExprStmt* stmt) {
...
 
+bool Semantics::IsIncluded(Decl* sym) {
+    const auto fileno = cc().sources()->GetSourceFileIndex(sym->pos());
+    if (fileno >= cc().sources()->opened_files().size()) { return false; }
+
+    return ! cc().sources()->opened_files()[fileno]->is_main_file();
+}
+
+template <typename DeclType>
+bool Semantics::IsIncludedStock(DeclType* sym) {
```
dvander, line 2318: This should be VarDeclBase, no template needed.

compiler/semantics.cpp
```diff
@@ -2307,6 +2307,18 @@ bool Semantics::CheckExprStmt(ExprStmt* stmt) {
     return true;
 }
 
+bool Semantics::IsIncluded(Decl* sym) {
+    const auto fileno = cc().sources()->GetSourceFileIndex(sym->pos());
```
dvander, line 2311: cc() should be cc_, here and elsewhere.

compiler/semantics.cpp
```diff
@@ -2307,6 +2307,18 @@ bool Semantics::CheckExprStmt(ExprStmt* stmt) {
...
+    if (fileno >= cc().sources()->opened_files().size()) { return false; }
+
+    return ! cc().sources()->opened_files()[fileno]->is_main_file();
```
dvander, line 2314: nit: "!cc_" (no space)

compiler/semantics.cpp
```diff
@@ -2307,6 +2307,18 @@ bool Semantics::CheckExprStmt(ExprStmt* stmt) {
     return true;
 }
 
+bool Semantics::IsIncluded(Decl* sym) {
+    const auto fileno = cc().sources()->GetSourceFileIndex(sym->pos());
+    if (fileno >= cc().sources()->opened_files().size()) { return false; }
```
dvander, line 2312: nit: no single-line conditionals. should be:

    if (fileno >= cc().sources()->opened_files().size()
        return false;

Though, I think this condition is unnecessary. We should never get a source index for a file not in the open list.

A1mDev:
The problem still exists with global variables, could you fix your PR?

dvander:
I'll fix up and land this.

PR #980, opened 2024-06-09
Update use of Removed ConfigParser.readfp() Function

c0rp3n:
# Solution

This was replaced with read_file() in version 3.2. Added a try cactch block around this code so we can delegate through to either version.

# Testing

Testing scripts runs correctly with python 3.12.4.


dvander:
Oh, Python.

PR #985, opened 2024-08-17
Make InstallBuiltinNatives() virtual.

NotnHeavy:
This PR does nothing except make `sp::PluginRuntime::InstallBuiltinNatives()` a virtual function, allowing it to be called in external libraries.

I'm not sure why this isn't already exposed - if there is a reason though, feel free to inform me!

### DIFF
--- a/vm/plugin-runtime.h
+++ b/vm/plugin-runtime.h
@@ -94,7 +94,7 @@ class PluginRuntime
   bool UsesHeapScopes();
 
   // Mark builtin natives as bound.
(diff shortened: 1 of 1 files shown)


dvander:
I don't think this does anything? It's not in `IPluginRuntime`, and nothing inherits from `PluginRuntime`.

NotnHeavy:
I thought this was related to SourcePawn's built-in float natives? Although I did take a quick glimpse - and I suppose that in the context of SourceMod, it uses its own float native implementations anyway.

PR #1018, opened 2025-01-05
Fix validation of rtti debug sections in VM

peace-maker:
When enabling the line debugger feature, the `.dbg.*` sections emitted by the current compiler don't pass the method-verifier in the VM anymore. Fix the problems and enable validation in tests too, to prevent this from diverging again.

This path can be triggered when enabling the line debugger in sourcemod. https://github.com/alliedmodders/sourcemod/pull/2240


tests/runtests.py
```diff
@@ -131,14 +131,14 @@ def find_shells(self):
       if rc == 0 and 'JIT' in stdout:
         self.shells.append({
           'path': path,
-          'args': [],
+          'args': ['--validate-debug-sections'],
```
dvander, line 134: Any reason not to make this the default for the shell?

PR #1019, opened 2025-01-05
Add functions to iterate function names to IPluginDebugInfo

peace-maker:
Allow to inspect the list of functions defined in a plugin as well as the corresponding source filename if available and requested.

I know the pre-RTTI debug information in the binaries should die, but I can't help but like the idea of being able to debug plugins compiled with old compilers too. We can remove that commit if it really isn't wanted.


vm/smx-v1-image.cpp
```diff
@@ -1060,6 +1060,92 @@ SmxV1Image::GetFileName(size_t index) const
...
+    if (sym->dimcount > 0)
+      cursor += sizeof(DimType) * sym->dimcount;
+    cursor += sizeof(SymbolType);
+  }
+  return func_count;
+}
+
+size_t
+SmxV1Image::NumFunctions() const {
+	if (rtti_methods_) {
```
dvander, line 1086: Mixed tabs/spaces

dvander (approved):

PR #1055, opened 2026-02-21
Add int64 support to SourcePawn.

dvander:
An int64 is internally allocated as two cells. That makes it kind of
like an array, except that arrays are not been able to participate in
any expressions beyond index operations. None of the arithmetic or logic
handling code is designed to work with arrays, so although "just
allocate two cells" is the obvious solution, it is not trivial.

As a result, this patch is fairly surgical and hardcodes int64 logic
into many touchpoints, rather than re-architecting the type system. This
was made possible by previous refactorings that made it easier to reason
about data flow in the compiler.

The huge effort to add an IR would have made this much easier. But that
effort is stalled due to... idiosyncracies of certain operations in
SourcePawn (like view_as, "any", implicit slices, etc). So while that
effort is shelved, here is what we did instead:

A local int64 allocates two local cells. When pushing an int64 onto the
operand stack, we push its address, as if it were an array or enum
struct. The weirdness is now that when using operators, we can't mutate
the address - it must be considered immutable. Thus we have to allocate
temporary variables. For example:

    int64 a = 10;
    int64 b = 20;
    int64 c = a + b;

Is internally deconstructed as:

    int a[2] = {10, 0};
    int b[2] = {20, 0};
    int __temp[2];
    AddInt64(__temp, a, b);
    int c[2] = __temp;

So, int64 will not be as efficient as int. To soften the blow we now
have int64 specific opcodes. There are no native calls for int64 use.

Why do we need temporaries? The alternative is to push/pop two cells on
the operand stack any time we interact with an int64. That's what the
JVM does. What we have, that the JVM doesn't, are virtual registers. Our
opcodes only operate on registers. The operand stack is just there to
save values that might be overwritten during complex expressions. It
would be a massive refactoring for us to generally support 64-bit
transfers between virtual registers and the operand stack. For example,
code that generically emits PUSH.PRI would have to know whether PRI
contains a 32-bit or 64-bit value. This is what the IR would have
solved.

What we're doing here is only modifying touchpoints where values are
actually consumed, instead of "every single place the operand stack is
used".

int64 is fully supported for enum struct fields, arrays, function
signatures (including natives), and all operators that apply to int.
Integer literals may now be 64-bit.

There are some restrictions on int that may be revisited if time
permits:
 - int64s are never compile-time constants, even as literals. They are
   not constant folded. Similarly int64s cannot be declared with "const".
   An exception is made for literals so global and array literals work.
 - int64 can't be used as an index to an array.
 - int64 can't be used in chained compares (a < b < c). Previously, all
   comparisons were internally handled as chained comparisons, which is
   very old, legacy code. For simplicity, binary comparisons are now
   emitted as binary expressions, which are simpler and more efficient.
   The chained comparison code needs to be refactored to support int64,
   and use of this feature is rare.

There are restrictions that won't change:
 - int64 cannot be used as an operator overload argument. Operator
   overloads will probably be removed in the near future.
 - int64 does not cast to or from "any", since they are different sizes.
 - The preprocessor does not support int64 expressions.
 - There are no implicit coercions from int64 to non-int types, like
   enums, methodmaps, bools, or floats. This is intentional. view_as
   must be used.

When writing a native that takes an int64, the params value contains an
address. The result of LocalToPhysAddr must be casted to an int64_t*.
When a native returns an int64_t, the address of the hidden stack slot
is stored in params[1]. Whatever is written to that address, is what the
plugin will see. The contents are NOT initialized to zero. The normal
C++ return value is ignored. An example of how to write an int64 native
is in vm/shell/shell.cpp, see "AddInt64".

There is a whole new test suite for int64, however, support is still
considered experimental so please report any bugs or any functional
difficulty using int64.


Fyren:
Something users might stumble over is doing `int64 a = 1 << 32` or `int64 a = hi << 32 | lo` (where `hi` is 32-bit).

Issue #1050, opened 2026-02-11
Function-like macros cause compilation error when immediately following an enum struct.


Localia-cn:
**Defining a macro with arguments and using it immediately after the `enum struct` causes the compiler to fail.**

```sourcepawn
#define INT_A1(%1) int A1;
enum struct S1
{
    int abc;
}
INT_A1(%1)  //error 155: expected newline, but found 'int'
```

**If the macro is defined without arguments, the code compiles successfully.**
```sourcepawn
#define INT_A1 int A1;

enum struct S1
{
    int abc;
}

// This compiles fine
INT_A1
```

Working Version: This issue is NOT present in SourceMod 1.10.
Broken Version: Verified as broken on SourceMod 1.13 (master).
> Note: I have not performed a bisect to identify the exact commit or version where the regression was introduced.

Issue #1035, opened 2025-06-16
Multiline macro functions are complaining about missing newline on SM compiler 1.13.0.7243 as well as 1.12.0.7201


naydef:
Code at [this link](https://github.com/Phil25/RTD/blob/26aba4fc19b7390fd1e4a8b341bb25abfae4e068/scripting/rtd/classes/rollers.sp#L36):
```
#define SET_PROP(%1,%2,%3) \
	public %1 Set%2(int client, %1 val){ \
		this.Set(client, val, %3);}
```

Output:
```
rtd/classes/rollers.sp(36) : error 155: expected newline, but found 'public'
    36 | #define SET_PROP(%1,%2,%3) \
------------------------------^

rtd/classes/rollers.sp(37) : error 155: expected newline, but found 'public'
    37 |         public %1 Set%2(int client, %1 val){ \
-------------------------------------^

rtd/classes/rollers.sp(39) : error 155: expected newline, but found 'public'
    39 | 
---------^
```
and continuing

This code was compiling correctly on SM 1.11 compiler

PR #1125, opened 2026-05-13
Fix error 155 when function-like macro follows enum struct / methodmap `}`

SCP-076:
https://github.com/alliedmodders/sourcepawn/issues/1035
https://github.com/alliedmodders/sourcepawn/issues/1050
A function-like macro right after a methodmap/enum struct/typeset `}` — or inside a methodmap body — gets error 155.  But these cases worked fine in 1.10.

The bug: `EnterMacro()` calls `lex()` to eat macro arguments like `(a)`. Those nested `lex()` calls write tokens into the ring buffer and bump the cursor, but `EnterMacro()` never cleaned up properly afterward.`peek_same_line()` then finds stale token data, thinks there's something on the same line, and fires error 155.

The fix:
- Save and restore cursor/depth/num_tokens in `EnterMacro()`, clear the polluted buffer slots.
- Snapshot the current token's line in `peek_same_line()` before `lex_tok()` can shift things around.

Note:  need more tests.

### Cases fixed
```sourcepawn

#define M(%1) int x_%1;


// ---- methodmap ----

methodmap MM {
    public MM() { return view_as<MM>(0); }
}
M(a)


// ---- enum struct ----

enum struct ES {
    int abc;
}
M(b)


// ---- typeset ----

typeset TS {
    function int (int a);
}
M(c)


// ---- struct (pstruct) ----

struct PS {
    public int abc;
}
M(d)


// ---- static_assert ----

static_assert(true)
M(e)

// ---- macro inside methodmap body ----

#define MM_FUNC(%1) public void %1(){}

methodmap ABC
{
    public void A1(){}
    MM_FUNC(A2)
}

```

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -2217,6 +2217,14 @@ int Lexer::peek_same_line() {
         return next_token()->id ? next_token()->id : tEOL;
     }
 
+    // Save the line of the current (last consumed) token before lexing ahead.
+    // lex_tok() may trigger macro expansion which calls nested lex() and
+    // advances the cursor by multiple positions. After lexpush(), the cursor
+    // may not point to the original token. We snapshot the line/file here
+    // so the comparison below uses the correct values.
+    uint32_t last_line = current_token()->start.line;
+    auto last_file = current_token()->start;
+
     // Make sure the next token is lexed, then buffer it.
     full_token_t next = lex_tok();
     if (next.id == 0 || next.id == tEOL)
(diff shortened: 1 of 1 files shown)


SCP-076:
Fix regression: EnterMacro() state restoration broke macro name shadowing by a local variable.[(l4d2_leaker_boomer_2768061.sp](https://github.com/dvander/sourcepawn-corpus/blame/ee1191647a125ae72f53c2cbff0af0b894ca612d/forums/N/o/r/NoroHime/l4d2_leaker_boomer_2768061.sp#L223))

Source Corpus Tests all passed now.

PR #1130, opened 2026-05-17
Expand CallArgs to make IForward easier to implement.

dvander:


include/sp_callable.h
```diff
@@ -48,15 +60,32 @@ struct CallArgs {
         argv[argc].u.addr = cell;
         argv[argc].flags = flags;
         argv[argc].type = ARG_CELL_BY_REF;
+        argv[argc].flags = 0;
```
Amaroq7, line 63: Is it correct here? Flags are already set in line 61

dvander: Good catch, fixed!

PR #1152, opened 2026-07-01
Expose `GetVersionString` in `ISourcePawnEnvironment`

Kenzzer:
For accelerator

### DIFF
--- a/include/sp_vm_api.h
+++ b/include/sp_vm_api.h
@@ -919,7 +919,6 @@ class ISourcePawnEnvironment
     // @brief Returns the code of the pending exception.
     virtual int GetPendingExceptionCode(const ExceptionHandler* handler) = 0;
 
-
     /**
      * @brief Returns the version string.
      *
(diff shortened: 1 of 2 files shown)


asherkin:
I vaguely recall that Accelerator only uses this as a proxy for the SourceMod version because it was easy to get - perhaps better to expose that more directly if that is the case?

Kenzzer:
> I vaguely recall that Accelerator only uses this as a proxy for the SourceMod version because it was easy to get - perhaps better to expose that more directly if that is the case?

Was going to be my original intention, until Dvander mentioned wanting to do a SP 2.0. If that's going to be an opt-in while running SM 1.13, then it makes sense to me accelerator dump should report SP version still, but if it's not an opt-in then yes we should expose a getter for SM version instead.

include/sp_vm_api.h
```diff
@@ -918,6 +918,14 @@ class ISourcePawnEnvironment
 
     // @brief Returns the code of the pending exception.
     virtual int GetPendingExceptionCode(const ExceptionHandler* handler) = 0;
+
```
dvander, line 921: nit: extra newline here

Kenzzer: I hope I can edit this with the web editor. Let's see...

PR #1153, opened 2026-07-03
Fix x64 switch tables when multiple cases share a target block

Mikusch:
The x64 JIT's dense switch jump tables breaks when several switch cases share one body (e.g. `case 1,2,3:`), since a PatchCodeLabel only supports a single pending use. Thread pending uses through the address table slots as a linked list so binding patches every referencing slot.


dvander:
The fix looks good, nice find. However PatchCodeLabel derives from CodeLabel base which explicitly says this behavior is not included. I'll fix that as a follow-up.

PR #1157, opened 2026-07-09
Implement float opcodes in the x64 JIT

Mikusch:
The x64 JIT currently stubs the legacy float opcodes. The compiler no longer emits these, but they still enter JIT-compiled code through float native remapping. The pcode reader rewrites `sysreq.n` calls to the natives in `sNativeMap` (e.g. `FloatCompare`, `float`, `FloatAdd`) into these opcodes, including for plugins that pass `SupportsPlugin`. When that happens there is no interpreter fallback. Compilation fails and the plugin cannot run. `FloatCompare` is a current, non-deprecated SourceMod API, so newly compiled plugins can hit this.

This PR ports the x86 SSE2 handlers and adds the missing memory-operand forms of `movss`, `addss`, `subss`, `mulss` and `divss` to the x64 assembler.

### DIFF
--- a/vm/x64/assembler-x64.h
+++ b/vm/x64/assembler-x64.h
@@ -863,14 +863,29 @@ class Assembler : public AssemblerBase
         emit3_sse(0x66, 0x0f, 0x6e, dest, src);
     }
 
+    void movss(FloatRegister dest, const Operand& src) {
+        emit3_sse(0xf3, 0x0f, 0x10, dest, src);
+    }
+    void addss(FloatRegister dest, const Operand& src) {
+        emit3_sse(0xf3, 0x0f, 0x58, dest, src);
+    }
     void addss(FloatRegister dest, FloatRegister src) {
         emit3_sse(0xf3, 0x0f, 0x58, dest, src);
+    }
+    void subss(FloatRegister dest, const Operand& src) {
+        emit3_sse(0xf3, 0x0f, 0x5c, dest, src);
     }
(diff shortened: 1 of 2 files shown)


dvander:
These are intentionally not implemented. Float native remapping is not supported in recompiled .smx files.

Mikusch:
> These are intentionally not implemented. Float native remapping is not supported in recompiled .smx files.

How are natives like `FloatCompare` supposed to work on x64 if these stay unimplemented? As far as I can tell the remapping applies to newly compiled files and the reader substitutes at `sysreq.n` for any bound, non-optional native. There's no check on the code version or feature mask.

If the intent is that the remapping shouldn't apply to modern binaries at all, we could gate `SetupFloatNativeRemapping` on the code feature mask instead. But that would change behavior on x86 too.

dvander:
Please attach a sample plugin that when compiled against the current compiler, fails to run on x64.

Mikusch:
> Please attach a sample plugin that when compiled against the current compiler, fails to run on x64.

```sourcepawn
#include <shell>

native int FloatCompare(float a, float b);

public void main()
{
    printnum(FloatCompare(2.0, 1.0));
}
```

[floatcompare-repro.zip](https://github.com/user-attachments/files/29938883/floatcompare-repro.zip)

Running this on master: `Error executing main: Unknown error code 0`

dvander:
Thanks, this should now be addressed on master w/ https://github.com/alliedmodders/sourcepawn/pull/1159

PR #1172, opened 2026-09-06
vm/v2: flatten heap-backed arrays in the variadic region of script calls

### LINKED ISSUE #1170: [Bug]: Garbled characters are output when forwarding char[] using any ... variable parameters (SourcePawn 2)
### Steps to Reproduce

```sourcepawn
#include <sourcemod>

public void OnPluginStart()
{
    char[] fmt = "(value1 %d) (str %s) (value2 %d)";
    char[] str = "some string";
    int value1 = 1234567;
    int value2 = 7654321;

    Func(fmt, value1, str, value2);
}

public void Func(char[] fmt, any ...)
{
    char buffer[1024];
    FormatEx(buffer, sizeof(buffer), fmt, ...);

    PrintToServer("[fmt %s]", fmt);
    PrintToServer("[buffer %s]", buffer);
}
```

### Expected Output
```
[fmt (value1 %d) (str %s) (value2 %d)]
[buffer (value1 1234567) (str some string) (value2 7654321)]
```

### Actual Output

```
[fmt (value1 %d) (str %s) (value2 %d)]
[buffer (value1 1234567) (str ) (value2 7654321)]
```

### Root causes analyzed by AI

The problem lies in the pcode lowering when the VM loads the plugin: LowerCall only "flattens" the heap array into a payload address when "parameters are directly passed to native". Parameters in the variable parameter area are packaged as is, causing the heap object reference to be passed directly to native's params[], and %s thus reads the header of SpArray as a string.

F1F88:
Resolved #1170  

Vararg cells are marshalled as addresses of flat data, which is what natives resolve through LocalToPhysAddr() / LocalToString(). A heap-backed array must be flattened first, otherwise the callee — or any native the variadic args are spread into — sees the SpArray header. Previously this was only done for direct native calls, so a heap char[] forwarded through a script variadic function reached the native as a raw heap ref and %s printed header bytes.

The guard is IsNonFlatArray(), not IsHeapItem(): that is the exact set of array types whose register value is a heap ref (Array / FixedArray / ArraySlice). FlatArray holds a data address and must be left alone, and Function / Object / Closure have no data field.

LL_ARRAY_TO_FLAT is an alias, not a copy, so mutation semantics and cost are unchanged. args_to_free is drained after EmitCall, so the array survives the call; and FreeReg is a no-op for the unowned VReg of an OP_LOAD_S local, so no spurious LL_RELEASE is emitted.

Note: This PR was primarily generated by AI.

### DIFF
--- a/tests/regressions/pass-heap-array-as-vararg.out
+++ b/tests/regressions/pass-heap-array-as-vararg.out
@@ -1,0 +1,1 @@
+(value1 1234567) (str some string) (value2 7654321)
--- a/tests/regressions/pass-heap-array-as-vararg.sp
+++ b/tests/regressions/pass-heap-array-as-vararg.sp
@@ -1,0 +1,16 @@
+#include <shell>
+
+void Func(char[] fmt, any ...)
+{
+    printf(fmt, ...);
+}
+
+public void main()
+{
+    char[] fmt = "(value1 %d) (str %s) (value2 %d)";
+    char[] str = "some string";
+    int value1 = 1234567;
+    int value2 = 7654321;
+
+    Func(fmt, value1, str, value2);
+}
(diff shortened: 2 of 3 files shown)


dvander:
Thanks for the analysis, this has been fixed via PR #1174.

